Issue #25, opened 2017-07-06
UNIX install script has `/tmp` hardcoded


euantorano:
This causes issues when installing on systems where access to `/tmp` is restricted (such as on shared web hosts that allow restricted SSH access). It should instead use the value of the `$TMPDIR` environment variable.

dom96:
Fixed by #26 :)

PR #26, opened 2017-07-06
#25 Use $TMPDIR environment variable when set

euantorano:
This will use the `$TMPDIR` environment variable when it is set, rather than the hardcoded `/tmp`. This should aid with portability and for instance lets me install Nim on a WebFaction server.


dom96:
Awesome, thanks. Remind me to deploy this if I don't do it within the next few days :)

euantorano:
> Remind me to deploy this if I don't do it within the next few days :)

One reminder as requested :smile:

dom96:
Thanks :)

Should be deployed now.

Issue #30, opened 2017-08-25
Redundant hint messages


dom96:
![screen shot 2017-08-25 at 14 23 54](https://user-images.githubusercontent.com/246651/29715884-15a47ab2-89a1-11e7-8a15-50989870d893.png)

Those hints are redundant when installing for the first time. Introduce a switch to disable them?

k0pernicus:
Maybe I can try to solve this issue, in order to learn more Nim and to contribute to the Nim community.
What is the "switch" you want to introduce for this?

dom96:
Yay. Awesome :)

Hrm, naming is tough.

Thinking of either ``--bootstrap`` or ``--first-install``. This won't be visible to the user so it doesn't matter too much, but it should be obvious to us. But yeah, pick whichever you like, I think I'm leaning towards ``--first-install``.

k0pernicus:
`--first-install` is great.
I'll try to submit a PR at the end of the afternoon.

dom96:
Fixed by #31.

PR #31, opened 2017-09-04
Solve #30 - Redundant hint messages

k0pernicus:
Add a simple change to avoid redundant hint message, installing Nim for the first time.
To do that, I added a new switch called 'firstInstall', that will enable a boolean value in `CliParams` (cliparams.nim).
This boolean value will be check in `switcher.nim`. If the condition is verified, the program will not print out hints about non-existing binaries in local path.

Simple resolution for #30

### DIFF
--- a/src/choosenim/cliparams.nim
+++ b/src/choosenim/cliparams.nim
@@ -57,8 +57,7 @@ Options:
                         installed. Default: ~/.choosenim.
   --nimbleDir:<dir>     Specify the Nimble directory where binaries will be
                         placed. Default: ~/.nimble.
-  --firstInstall        Specify user friendly and less redundant
-                        messages. Used by install script.
+  --firstInstall        Used by install script.
 """
 
 proc command*(params: CliParams): string =
--- a/src/choosenim/cliparams.nim
+++ b/src/choosenim/cliparams.nim
@@ -9,6 +9,7 @@ type
   CliParams* = ref object
     commands*: seq[string]
     choosenimDir*: string
+    firstInstall*: bool
     nimbleOptions*: Options
 
 let doc = """
@@ -56,6 +57,10 @@ Options:
                         installed. Default: ~/.choosenim.
   --nimbleDir:<dir>     Specify the Nimble directory where binaries will be
                         placed. Default: ~/.nimble.
+  --firstInstall       Specify you are trying to install Nim for the first
+                        time. This option supports user friendly messages to
+                        understand how choosenim is working, and less redundant
+                        messages.
 """
 
 proc command*(params: CliParams): string =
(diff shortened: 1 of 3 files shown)


src/choosenim/cliparams.nim
```diff
@@ -56,6 +57,10 @@ Options:
                         installed. Default: ~/.choosenim.
   --nimbleDir:<dir>     Specify the Nimble directory where binaries will be
                         placed. Default: ~/.nimble.
+  --firstInstall       Specify you are trying to install Nim for the first
```
dom96: Please shorten this and make sure it's aligned with the rest of the help (you're missing a single space before ``Specify`` :))

This shouldn't be used by users, so just write "Used by install script."

src/choosenim/switcher.nim
```diff
@@ -111,8 +111,11 @@ proc writeProxy(bin: string, params: CliParams) =
   # Check whether this is in the user's PATH.
   let fromPATH = findExe(bin)
   if fromPATH == "":
-    display("Hint:", "Binary '$1' isn't in your PATH. Add '$2' to your PATH." %
-            [bin, params.getBinDir()], Warning, HighPriority)
+    # If the binary does not exists in the bin directory, and the option
+    # firstInstall is not set, display an hint to indicate the solution.
+    if not params.firstInstall:
```
dom96: Reduce this nesting and remove the comment (the code is self-explanatory anyway).

src/choosenim/cliparams.nim
k0pernicus: Done!

src/choosenim/switcher.nim
k0pernicus: Done!

k0pernicus:
@dom96 I updated the PR ;-)
Actually, I am not convinced about the nesting reduction but...

src/choosenim/cliparams.nim
```diff
@@ -56,6 +57,8 @@ Options:
                         installed. Default: ~/.choosenim.
   --nimbleDir:<dir>     Specify the Nimble directory where binaries will be
                         placed. Default: ~/.nimble.
+  --firstInstall        Specify user friendly and less redundant
```
dom96: I think it's enough to just write "Used by install script" here. No need for the user to know what this does.

dom96:
> Actually, I am not convinced about the nesting reduction but...

It is a shame that you need to edit the ``elif`` condition, but I do still prefer it.

k0pernicus:
No problem :-)

src/choosenim/cliparams.nim
k0pernicus: Done!

dom96:
Brilliant, thank you!

Issue #41, opened 2017-10-31
Choosenim doesn't support proxies


genotrance:
Running choosenim behind a proxy server isn't supported. This is a common situation for developers and important to support.

PR #48, opened 2018-01-09
Fix for #41 - proxy support

genotrance:
Sourced getProxy() from Nimble

### DIFF
--- a/src/choosenim/download.nim
+++ b/src/choosenim/download.nim
@@ -1,4 +1,4 @@
-import httpclient, strutils, os, terminal, times, math
+import httpclient, strutils, os, terminal, times, math, uri
 
 import nimblepkg/[version, cli]
 when defined(curl):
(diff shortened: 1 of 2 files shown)


src/choosenim/download.nim
```diff
@@ -138,8 +138,31 @@ when defined(curl):
...
+        getCurrentExceptionMsg(), Warning, HighPriority)
+
+  if url.len > 0:
+    var parsed = parseUri(url)
+    if parsed.scheme.len == 0 or parsed.hostname.len == 0:
+      parsed = parseUri("http://" & url)
+    let auth =
+      if parsed.username.len > 0: parsed.username & ":" & parsed.password
+      else: ""
+    return newProxy($parsed, auth)
```
dom96: I find this code odd. Doesn't ``httpclient`` provide this logic? If not, I think it should. But we can take care of that later.

genotrance: Nope, this isn't in httpclient.nim yet. I know you mentioned this separately as well but I added this to choosenim for the near term.

genotrance: Nope, this isn't in httpclient.nim yet. I know you mentioned this separately as well but I added this to choosenim for the near term.

dom96:
This will do for now but I would like it in the stdlib.

PR #62, opened 2018-05-24
Fixes for 0.18.1

genotrance:
Fixes for ""[0]

### DIFF
--- a/src/choosenim/download.nim
+++ b/src/choosenim/download.nim
@@ -34,7 +34,10 @@ proc showIndeterminateBar(progress, speed: BiggestInt, lastPos: var int) =
   stdout.flushFile()
 
 proc showBar(fraction: float, speed: BiggestInt) =
-  eraseLine()
+  try:
+    eraseLine()
+  except OSError:
+    discard
   let hashes = repeat('#', int(fraction * progressBarLength))
   let spaces = repeat(' ', progressBarLength - hashes.len)
   stdout.write("[$1$2] $3% $4kb/s" % [
--- a/tests/tester.nim
+++ b/tests/tester.nim
@@ -42,21 +42,21 @@ proc outputReader(stream: Stream, missedEscape: var bool): string =
   while true:
     let c = stream.readStr(1)
 
-    if c.len() != 0:
-      case c[0]
-      of '\c', '\l':
-        result.add(c[0])
-        return
-      of '\27':
-        if result.len > 0:
-          missedEscape = true
-          return
-
-        handleEscape()
-      else:
-        result.add(c[0])
-    else:
+    if c.len() == 0:
       return
+
+    case c[0]
+    of '\c', '\l':
+      result.add(c[0])
+      return
+    of '\27':
+      if result.len > 0:
+        missedEscape = true
+        return
+
+      handleEscape()
+    else:
+      result.add(c[0])
 
 proc exec(args: varargs[string], exe=exePath,
           yes=true, liveOutput=false,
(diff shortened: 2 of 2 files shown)


tests/tester.nim
```diff
@@ -42,18 +42,21 @@ proc outputReader(stream: Stream, missedEscape: var bool): string =
...
+        if result.len > 0:
+          missedEscape = true
+          return
+
+        handleEscape()
+      else:
+        result.add(c[0])
     else:
-      result.add(c[0])
+      return
```
dom96: When writing code like this, instead of wrapping the code in an if, just write the following above the code:

```nim
if c.len() == 0:
  return
```

It's the same and it makes the code less messy.

src/choosenim/download.nim
```diff
@@ -34,7 +34,10 @@ proc showIndeterminateBar(progress, speed: BiggestInt, lastPos: var int) =
   stdout.flushFile()
 
 proc showBar(fraction: float, speed: BiggestInt) =
-  eraseLine()
+  try:
+    eraseLine()
+  except OSError:
+    discard
```
dom96: This isn't a valid fix for whatever issue this fixes.

tests/tester.nim
```diff
@@ -68,7 +71,9 @@ proc exec(args: varargs[string], exe=exePath,
       quotedArgs.add("--chooseNimDir:" & choosenimDir)
   quotedArgs.add("--noColor")
 
-  quotedArgs = quoted_args.map((x: string) => ("\"" & x & "\""))
+  for i in 0..quotedArgs.len-1:
+    if " " in quotedArgs[i]:
+      quotedArgs[i] = "\"" & quotedArgs[i] & "\""
```
dom96: I don't see a reason to do this. Quoting everything is fine, and the code is much more succinct.

src/choosenim/download.nim
genotrance: This happens when running tests - there's no terminal available so `eraseLine()` raises OSError.

tests/tester.nim
genotrance: Yes, but it is failing tests even on 0.18.0 since it is appending spaces or \" to file paths for --nimbleDir or --choosenimDir which then no longer resolve as paths. It's easier fixing the tester than changing choosenim `parseCliParams()`.

tests/tester.nim
genotrance: Updated.

genotrance:
Verified tests pass on latest #head of Nim on both Windows and Linux. I require the patched `untar` without which the following error occurs.

```
 Extracting nim-0.16.0.tar.xz
       Tip: 1 messages have been suppressed, use --verbose to show them.
     Error: choosenim.nim(206)       choosenim
        ... choosenim.nim(187)       performAction
        ... choosenim.nim(89)        choose
        ... choosenim.nim(72)        chooseVersion
        ... choosenim.nim(39)        installVersion
        ... utils.nim(51)            extract
        ... Unable to extract. Error was 'invalid integer: '.
```

Invalid integer is raised by `parseOctInt()`. I'm checking to see if there were any code changes in that proc recently.

src/choosenim/download.nim
dom96: Then how have the tests worked before?

tests/tester.nim
dom96: Again, how did this pass before?

genotrance: It didn't pass for me even on 0.18.0, I think it is probably because of running the tests on Windows. It never passed for me because the command line params were getting messed up.

src/choosenim/download.nim
genotrance: Only Windows raises OSError, Linux does not so I can only guess that this wasn't tested on Windows before.

PR #75, opened 2018-08-08
Remove / prefix when calling choosenim binary

euantorano:
`$temp_prefix` defaults to `/tmp` (or uses the `$TMPDIR` environment variable). Prepending the `/` prefix results in trying to call a path like `//tmp/choosenim-0.3.2_linux_amd64`.

### DIFF
--- a/scripts/choosenim-unix-init.sh
+++ b/scripts/choosenim-unix-init.sh
@@ -49,10 +49,10 @@ install() {
     fi
 
     # Install Nim from desired channel.
-    "/$temp_prefix/$filename" $CHOOSE_VERSION --firstInstall < /dev/tty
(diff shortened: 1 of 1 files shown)


dom96:
this doesn't actually cause any issues, though, right? The default value for this var seems to be ``/tmp`` unless my bash is failing me

euantorano:
@dom96 It might if `$TMPDIR` is set to something like `./` for some reason as the script would then try to call `/./` which obviously isn't correct. The default vlaue being `/tmp` means is calls `//tmp/choosenim-0.3.2_linux_amd64` as noted in the first comment.

genotrance:
This sounds fair and can be pulled in.

PR #82, opened 2018-08-17
Improved handling of proxy environment variables

trialism:
Environment variables only use capital letters and `_`, `http_proxy` is not enough.


dom96:
Does setting ``http_proxy`` work for you then?

trialism:
> Does setting http_proxy work for you then?

Yes, but env vars are always in caps.

dom96:
Okay, thanks for testing.

trialism:
Thanks for the merge!

PR #83, opened 2018-08-18
added windows 64bit support

emekoi:
now choosenim can build 64 bit nim on windows.

### DIFF
--- a/src/choosenim/builder.nim
+++ b/src/choosenim/builder.nim
@@ -7,8 +7,10 @@ import cliparams, download, utils, common, telemetry
 
 proc buildFromCSources() =
   when defined(windows):
-    doCmdRaw("build.bat")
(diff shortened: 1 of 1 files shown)


dom96:
It's not that simple :)

emekoi:
true.

emekoi:
@dom96?

dom96:
Because choosenim installs a 32bit MINGW by the default

emekoi:
so i would just need to add support for download 64bit mingw?

dom96:
This PR already handles this too: https://github.com/dom96/choosenim/pull/63

emekoi:
i guess i'll close this then.

dom96:
Sorry :) Thank you for giving this a shot.

PR #89, opened 2018-09-29
'choosenim installed' shows Nim versions located in the toolchain folder

ThomasTJdev:
See all available Nim versions located in the toolchain folder.

### DIFF
--- a/src/choosenim.nim
+++ b/src/choosenim.nim
@@ -180,7 +180,7 @@ proc show(params: CliParams) =
     versions.add(versionAvailable)
 
   if versions.len() > 1:
-    versions.sort(order=Descending)
+    versions.sort(system.cmp, Descending)
     if versions.contains("#head"):
       versions.del(find(versions, "#head"))
       versions.insert("#head", 0)


dom96:
I think this should simply be merged into ``show``, we don't need a new command for this. Please make the output ``git branch``-like.

ThomasTJdev:
Thank you for feedback! I'll update the PR.

@genotrance, I really like the the idea of `list all available`, but that deserves a separate PR.

ThomasTJdev:
**Ver 2** is now available in the PR.

genotrance:
Given the currently selected version is already mentioned at the top, I don't think there's a need to show it again in the list below and highlighting it. Perhaps we need to update that output to state that it is the selected version?

It will also be nice to have some title for the list of selectable versions:

```
Available versions:
   0.19.0
   0.18.0
   #hash-truncated1
   #hash-truncated2...
```

dom96:
I think it would be good to remove the "Version: ..." and add "Versions:\n" above the version list in place of it.

ThomasTJdev:
I personally prefer **V3** even though it is redundant. I like the git branch look, but I think the current `choosenim show` view is more **smooth**.

### V3
choosenim_v7

### V4
choosenim_v5

### V5
choosenim_v6

dom96:
v3 looks good

narimiran:
> v3 looks good

Can this be merged, then?

Issue #88, opened 2018-09-28
[Feature Request] Some sort of `--force` argument for `choosenim update ...`


chrisokuda:
There is no current (known) method to force `choosenim` to fully redownload the sources and rebuild a toolchain.

dom96:
You can ``rm`` the toolchain in ~/.choosenim/toolchains

mcintyrechris:
Sure, but a `--force` switch would be nice, too.

mcintyrechris:
I went ahead and implemented it in a fairly elegant way in PR #91.

genotrance:
What's the benefit of this feature? Choosenim already allows updating [special installs](https://github.com/dom96/choosenim/blob/master/src/choosenim/channel.nim#L49) from git since those can be expected to change. It only blocks for stable versioned installs which are fixed anyway.

Only reasons would be if install failed for some reason which would be a separate bug, or if gcc compiler was changed.

genotrance:
Third reason perhaps is to reinstall based on updates to choosenim - for example, I'm planning a PR to fix #98 which would delete c_code directory saving space. You could simply do `update --force` for an installed version and get the benefit of that fix. Of course, it isn't the best example since you can just rm -rf the directory but is an example nonetheless.

A marginally better example is PR #63 where an existing self-compiled install can be replaced with the official Nim binary ZIP for Windows.

chrisokuda:
nevermind...

PR #91, opened 2018-10-06
[Feature Addition] Add force update switch `--force`

chrisokuda:
This PR adds a switch `--force` (Issue #88) which causes `choosenim` to always delete and re-download, rebuild, etc. the entire toolchain for whichever version/channel is to be updated. This does nothing for any command other than `update`.

**Example:** `choosenim update stable --force`

_I also fixed a minor `.gitignore` capitalization mistake._

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@ src/choosenim/proxyexe
 *.exe
 
 tests/nimcache
-tests/chooseNimDir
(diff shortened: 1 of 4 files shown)


genotrance:
Discussion in issue #88

chrisokuda:
nevermind...

PR #95, opened 2018-10-12
Add nimpretty to the list of installed tools.

moigagoo:


moigagoo:
Note that it's not a pure PR as it does other stuff than advertised. I had to tweak onProgressChanged in order to compile choosenim, and I switched from parseopt2 to parseopt to eliminate the deprecation warning.

dom96:
Thanks!

dom96:
I don't have time to do a release right now.

Issue #96, opened 2018-10-17
Choosenim crashes when built with 0.18.0 and 0.19.0


genotrance:
With 0.18.0
```
> bin/choosenim --nimbleDir:tests/nimbleDir --chooseNimDir:tests/choosenimDir /tmp/asd

Traceback (most recent call last)
choosenim.nim(211)       choosenim
telemetry.nim(170)       report
asyncmacro.nim(401)      reportException
asyncmacro.nim(34)       reportException_continue
analytics.nim(108)       reportExceptionIter
asyncmacro.nim(401)      postContent
asyncmacro.nim(34)       postContent_continue
httpclient.nim(1286)     postContentIter
asyncmacro.nim(401)      post
asyncmacro.nim(34)       post_continue
httpclient.nim(1265)     postIter
asyncmacro.nim(401)      requestAux
asyncmacro.nim(37)       requestAux_continue
httpclient.nim(1175)     requestAuxIter
asyncmacro.nim(401)      send
asyncmacro.nim(34)       send_continue
asyncnet.nim(435)        sendIter
asyncmacro.nim(401)      sendPendingSslData
asyncmacro.nim(34)       sendPendingSslData_continue
asyncnet.nim(202)        sendPendingSslDataIter
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
```

With 0.19.0
```
> bin/choosenim --nimbleDir:tests/nimbleDir --chooseNimDir:tests/choosenimDir /tmp/asd

Traceback (most recent call last)
choosenim.nim(211)       choosenim
telemetry.nim(170)       report
asyncmacro.nim(307)      reportException
asyncmacro.nim(36)       reportException_continue
analytics.nim(108)       reportExceptionIter
asyncmacro.nim(307)      postContent
asyncmacro.nim(36)       postContent_continue
httpclient.nim(1294)     postContentIter
asyncmacro.nim(307)      post
asyncmacro.nim(36)       post_continue
httpclient.nim(1273)     postIter
asyncmacro.nim(307)      requestAux
asyncmacro.nim(39)       requestAux_continue
httpclient.nim(1183)     requestAuxIter
asyncmacro.nim(307)      send
asyncmacro.nim(36)       send_continue
asyncnet.nim(241)        sendIter
asyncmacro.nim(307)      sendPendingSslData
asyncmacro.nim(36)       sendPendingSslData_continue
asyncnet.nim(200)        sendPendingSslDataIter
[...]

PR #97, opened 2018-10-18
Issue #96 - Add analytics proxy support

genotrance:
- Move `getProxy()` to utils.nim since download.nim and telemetry.nim both need it
- Sent proxy info to `newAsyncAnalytics()`


genotrance:
Build failed since I bumped analytics dependency to 0.1.1 which is another PR that isn't accepted yet.

https://github.com/dom96/analytics/pull/1

Once that is accepted and tagged, this will build fine.

PR #100, opened 2019-01-03
Fix issue #16 and #27

genotrance:
Detect if 64-bit and run build64.bat for csources. Fixes issue #16.


genotrance:
Also fixes #27.

genotrance:
Once this PR is accepted, I can create a PR for installing Windows binary builds. Need the gcc detection logic to implement that feature.

PR #102, opened 2019-01-08
Install windows binary releases

genotrance:
Much cleaner than before

### DIFF
--- a/src/choosenim.nim
+++ b/src/choosenim.nim
@@ -30,13 +30,24 @@ proc parseVersion(versionStr: string): Version =
   result = newVersion(versionStr)
 
 proc installVersion(version: Version, params: CliParams) =
+  when defined(windows):
+    # Add MingW bin dir to PATH so download step can find gcc.
+    if findExe("gcc") == "" and dirExists(params.getMingwBin()):
+      let pathEnv = getEnv("PATH")
+      putEnv("PATH", params.getMingwBin() & PathSep & pathEnv)
+      defer:
+        putEnv("PATH", pathEnv)
+
   # Install the requested version.
   let path = download(version, params)
   # Extract the downloaded file.
   let extractDir = params.getInstallationDir(version)
   # Make sure no stale files from previous installation exist.
   removeDir(extractDir)
-  extract(path, extractDir)
+  if path.splitFile().ext == ".zip":
+    extract(path, extractDir.parentDir())
+  else:
+    extract(path, extractDir)
   # A "special" version is downloaded from GitHub and thus needs a `.git`
   # directory in order to let `koch` know that it should download a "devel"
   # Nimble.
--- a/src/choosenim.nim
+++ b/src/choosenim.nim
@@ -33,15 +33,15 @@ proc installVersion(version: Version, params: CliParams) =
   # Install the requested version.
   let path = download(version, params)
   # Extract the downloaded file.
-  let extractDir =
-    if ("nim-$1_x" % $version) in path.splitFile().name:
-      # Extracting a Windows binary release ZIP which already has a
-      # nim-$version directory so extract to base install directory
-      params.getInstallDir()
-    else:
-      params.getInstallationDir(version)
+  let extractDir = params.getInstallationDir(version)
   # Make sure no stale files from previous installation exist.
   removeDir(extractDir)
+  if ("nim-$1_x" % $version) in path.splitFile().name:
+    # Extracting a Windows binary release ZIP which already has a
+    # nim-$version directory so extract to base install directory
+    extract(path, params.getInstallDir())
+  else:
+    extract(path, extractDir)
   # A "special" version is downloaded from GitHub and thus needs a `.git`
   # directory in order to let `koch` know that it should download a "devel"
   # Nimble.
--- a/src/choosenim.nim
+++ b/src/choosenim.nim
@@ -30,24 +30,18 @@ proc parseVersion(versionStr: string): Version =
   result = newVersion(versionStr)
 
 proc installVersion(version: Version, params: CliParams) =
-  when defined(windows):
-    # Add MingW bin dir to PATH so download step can find gcc.
-    if findExe("gcc") == "" and dirExists(params.getMingwBin()):
-      let pathEnv = getEnv("PATH")
-      putEnv("PATH", params.getMingwBin() & PathSep & pathEnv)
-      defer:
-        putEnv("PATH", pathEnv)
-
   # Install the requested version.
   let path = download(version, params)
   # Extract the downloaded file.
-  let extractDir = params.getInstallationDir(version)
+  let extractDir =
+    if ("nim-$1_x" % $version) in path.splitFile().name:
+      # Extracting a Windows binary release ZIP which already has a
+      # nim-$version directory so extract to base install directory
+      params.getInstallDir()
+    else:
+      params.getInstallationDir(version)
   # Make sure no stale files from previous installation exist.
   removeDir(extractDir)
-  if path.splitFile().ext == ".zip":
-    extract(path, extractDir.parentDir())
-  else:
-    extract(path, extractDir)
   # A "special" version is downloaded from GitHub and thus needs a `.git`
   # directory in order to let `koch` know that it should download a "devel"
   # Nimble.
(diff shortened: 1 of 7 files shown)


src/choosenim.nim
```diff
@@ -30,13 +30,24 @@ proc parseVersion(versionStr: string): Version =
...
+
   # Install the requested version.
   let path = download(version, params)
   # Extract the downloaded file.
   let extractDir = params.getInstallationDir(version)
   # Make sure no stale files from previous installation exist.
   removeDir(extractDir)
-  extract(path, extractDir)
+  if path.splitFile().ext == ".zip":
+    extract(path, extractDir.parentDir())
```
dom96: `extract` shouldn't have different behaviour based on the archive you're extracting, this if branch should be in the `extract`.

src/choosenim.nim
```diff
@@ -30,13 +30,24 @@ proc parseVersion(versionStr: string): Version =
   result = newVersion(versionStr)
 
 proc installVersion(version: Version, params: CliParams) =
+  when defined(windows):
+    # Add MingW bin dir to PATH so download step can find gcc.
+    if findExe("gcc") == "" and dirExists(params.getMingwBin()):
+      let pathEnv = getEnv("PATH")
+      putEnv("PATH", params.getMingwBin() & PathSep & pathEnv)
+      defer:
+        putEnv("PATH", pathEnv)
```
dom96: Please move this to the appropriate place in ``download`` (it should actually go into ``getGccArch`` I think?).

dom96: Even if it means code duplication in ``build``.

src/choosenim/download.nim
```diff
@@ -217,8 +219,21 @@ proc downloadImpl(version: Version, params: CliParams): string =
...
+      url = winBinaryZipUrl % [$version, $arch]
+      if not needsDownload(params, url, outputPath): return outputPath
+      try:
+        downloadFile(url, outputPath, params)
+        return outputPath
+      except HttpRequestError:
+        display("Info:", "Binary build unavailable, building from source",
+                priority = HighPriority)
+
+    url = websiteUrl % $version
```
dom96:
```suggestion
```

src/choosenim/download.nim
```diff
@@ -217,8 +219,21 @@ proc downloadImpl(version: Version, params: CliParams): string =
   else:
     display("Downloading", "Nim $1 from $2" % [$version, "nim-lang.org"],
             priority = HighPriority)
-    let url = websiteUrl % $version
-    var outputPath: string
+
+    var
+      url: string
+      outputPath: string
```
dom96: Move these definitions below the `when`

src/choosenim/download.nim
```diff
@@ -217,8 +219,21 @@ proc downloadImpl(version: Version, params: CliParams): string =
...
+    when defined(Windows):
+      url = winBinaryZipUrl % [$version, $arch]
```
dom96:
```suggestion
      let winUrl = winBinaryZipUrl % [$version, $arch]
```

src/choosenim/download.nim
```diff
@@ -217,8 +219,21 @@ proc downloadImpl(version: Version, params: CliParams): string =
...
+      if not needsDownload(params, url, outputPath): return outputPath
```
dom96:
```suggestion
      if not needsDownload(params, winUrl, outputPath): return outputPath
```

src/choosenim/download.nim
```diff
@@ -217,8 +219,21 @@ proc downloadImpl(version: Version, params: CliParams): string =
...
+      try:
+        downloadFile(url, outputPath, params)
```
dom96:
```suggestion
        downloadFile(winUrl, outputPath, params)
```

src/choosenim/download.nim
```diff
@@ -217,8 +219,21 @@ proc downloadImpl(version: Version, params: CliParams): string =
   else:
     display("Downloading", "Nim $1 from $2" % [$version, "nim-lang.org"],
             priority = HighPriority)
-    let url = websiteUrl % $version
-    var outputPath: string
+
+    var
+      url: string
+      outputPath: string
+    when defined(Windows):
```
dom96: It seems like it would be relatively easy to check for the presence of `powershell` on the system so that we can continue to support Windows XP. Please do this by either:

* Using the ``osinfo`` package
* Running `powershell -v` (guessing the syntax here, just something that will always work on systems with powershell) and verifying it succeeds

If powershell isn't supported then we want to build from source.

src/choosenim.nim
genotrance: This is only because the zip file contents already has a `nim-$version` folder in it. Same will apply to xz files as well when we migrate to nimarchive. I can move it to `extract()` if preferred though.

genotrance:
I have addressed all your concerns - PTAL.

BTW, CI ran fine on my [instance](https://travis-ci.org/genotrance/choosenim/builds/479256174) of Travis but failed here for some reason.

genotrance:
Okay passed this time - looks like tester has some issue in detecting successful pass.

src/choosenim/switcher.nim
```diff
@@ -42,8 +42,12 @@ proc areProxiesInstalled(params: CliParams, proxies: openarray[string]): bool =
     if contents != proxyExe:
       return false
 
-proc isCCInPath(params: CliParams): bool =
-  return findExe("gcc") != "" or findExe("clang") != ""
+proc isCCInPath*(params: CliParams): bool =
+  when defined(Windows):
+    return findExe("gcc") != ""
+  else:
+    # Fix issue #104
```
dom96: This statement is ambiguous, is it a TODO or a "This code fixes issue #104"?

I'm assuming the former, but you seem to already be sort of fixing it? Just fix it properly.

* For Windows: check for gcc only
* For macOS (or freebsd): check for clang only
* For Linux: check for gcc only

And maybe rename this function to `isDefaultCCInPath` since these are the default C compilers used by Nim on these platforms.

src/choosenim/builder.nim
```diff
@@ -82,7 +82,9 @@ proc build*(extractDir: string, version: Version, params: CliParams) =
   setCurrentDir(extractDir)
   # Add MingW bin dir to PATH so that `build.bat` script can find gcc.
   let pathEnv = getEnv("PATH")
-  putEnv("PATH", params.getMingwBin() & PathSep & pathEnv)
+  when defined(windows):
+    if not isCCInPath(params) and dirExists(params.getMingwBin()):
```
dom96: Why is this necessary?

src/choosenim/download.nim
```diff
@@ -199,6 +200,15 @@ proc needsDownload(params: CliParams, downloadUrl: string,
     return false
 
 proc downloadImpl(version: Version, params: CliParams): string =
+  # Add MingW bin dir to PATH so getGccArch can find gcc.
```
dom96: without this ``getGccArch`` won't work, so this code should be inside ``getGccArch``

src/choosenim.nim
```diff
@@ -36,7 +36,12 @@ proc installVersion(version: Version, params: CliParams) =
   let extractDir = params.getInstallationDir(version)
   # Make sure no stale files from previous installation exist.
   removeDir(extractDir)
-  extract(path, extractDir)
+  if ("nim-$1_x" % $version) in path.splitFile().name:
```
dom96: Pretty sure this will break installation on other platforms. Also, my ``untar`` library already does this, so your new proc ``extractZip`` should do the same.

See how I do it in ``untar`` and do the same thing: https://github.com/dom96/untar/blob/master/src/untar.nim#L127 (also note that ``-j`` can be used for ``unzip``)

src/choosenim/switcher.nim
genotrance: I was fixing it based on our IRC conversation. I've updated it.

src/choosenim/builder.nim
genotrance: It is consistent with the other instance. And it is only needed on Windows so this makes it clear.

src/choosenim/download.nim
genotrance: I put it in `downloadImpl()` to avoid having to propagate `params` all over the place and minimize changes. I've changed it again.

src/choosenim.nim
genotrance: Note that `-j` will get rid of all directory structure, not just the base dir. It will extract all files into the destination as one big blob.

I'll do this in extractZip().

PR #105, opened 2019-01-14
Enable init.sh for git bash on Windows / Travis

genotrance:


scripts/choosenim-unix-init.sh
```diff
@@ -57,11 +67,20 @@ install() {
 
   # Copy choosenim binary to Nimble bin.
   local nimbleBinDir=`"$temp_prefix/$filename" --getNimbleBin`
-  cp "$temp_prefix/$filename" "$nimbleBinDir/choosenim"
+  if [ "$platform" = "windows_i386" ]; then
+    # Copy choosenim.exe and DLL files for Windows
+    cp "$temp_prefix/$filename" "$nimbleBinDir/."
+    cp "$temp_prefix/choosenim/libeay32.dll" "$nimbleBinDir/."
+    cp "$temp_prefix/choosenim/ssleay32.dll" "$nimbleBinDir/."
```
dom96: I think a better option would be to just run the .bat file that's in the zip.

genotrance: I thought of that but the bat file doesn't copy the DLL files over, nor is it in source control to improve.

dom96: Something must be copying it, maybe it's actually choosenim?

genotrance: Well, choosenim downloads DLLs.zip anyway because of `needsDLLInstall()`. The DLLs are probably included in the choosenim ZIP just to run the first install of Nim after which DLLs will be present in the bin directory. No use copying them over.

genotrance:
PTAL

genotrance:
@dom96 do you need anything else in this PR?

dom96:
Seems good. Thanks.

PR #107, opened 2019-01-14
Enable xz when unxz is available

genotrance:


dom96:
Retrying build, is it flaky?

genotrance:
Let's close this for the near term. Once I get nimarchive working with nimterop, we can pull that into choosenim and extract xz files natively.

Issue #22, opened 2017-06-28
Add command to list available versions


samdmarshall:
There should be a subcommand `list` that lists out the available released versions so we don't have to guess at what is and isn't available (like, why `0.16.0` works but `0.16.1` does not).

k0pernicus:
Maybe a local `list` command would be useful too, right?
I can work on this issue and propose a solution, if it's okay for you @dom96

dom96:
@k0pernicus feel free to work anything you like. I may request some changes in PRs though :)

samdmarshall:
has any work been made on this? it is something i could work on and submit a PR for?

dom96:
go for it

samdmarshall:
i'll give it my best; i cannot guarantee i can completely follow-through with this.

PR #112, opened 2019-03-18
adding command `versions` to list installable versions of nim

samdmarshall:
Overview:
i threw this together today in an attempt fix an outstanding issue i've had (#22), namely: being able to list which versions of Nim can be installed by `choosenim`.

Important Notes:
* Added a new command named `versions`, that takes an optional flag `--installed` which limits displayed output to only versions that are currently installed (avoids performing network requests as well).
* Added new code to query against the github releases of Nim, to give names/numbers to the versions that it was already able to fetch.
* Matched the display style of the "active version" to that shown by the `show` command.
* Added method `normalizeVersion` in `versions.nim` to try to standardize the what version numbers are displayed (removing all prefixed "#" and "v", so it only shows the semver number). I hope this can be removed at some point where a more standard way of displaying the version numbers can be put in place.

### DIFF
--- a/src/choosenim.nim
+++ b/src/choosenim.nim
@@ -211,24 +211,36 @@ proc list(params: CliParams) =
     else: getAvailableVersions(params)
 
   #[ Display version information,now that it has been collected ]#
+  proc isActiveTag(params: CliParams, version: string): string =
+    let tag = if version == currentVersion: "*"
+              else: " "
+    return tag
+  proc isLatestTag(params: CliParams, version: string): string =
+    let tag = if isLatestVersion(params, version): " (latest)"
+            else: ""
+    return tag
+  proc canUpdateTag(params: CliParams, channel: string): string =
+    let version = getChannelVersion(channel, params, live = true)
+    let channelVersion = parseVersion(version)
+    let tag = if canUpdate(channelVersion, params): " (update available!)"
+              else: ""
+    return tag
 
-  # display the currently selected version info
-  display("Current:", "", priority = HighPriority)
-  display("  Version:", currentVersion & isLatestTag(params, currentVersion), priority = HighPriority)
+
   if currentChannel.len > 0:
-    display("  Channel:", currentChannel, priority = HighPriority)
-  echo ""
+    display("Channel:", currentChannel & canUpdateTag(params, currentChannel), priority = HighPriority)
+    echo ""
 
   # local versions
   display("Installed:", " ", priority = HighPriority)
   for version in localVersions:
-    display("", version & isLatestTag(params, version), priority = HighPriority)
+    display(isActiveTag(params, version), version & isLatestTag(params, version), priority = HighPriority)
   for version in specialVersions:
-    display("", version, priority = HighPriority)
+    display(isActiveTag(params, version), version, priority = HighPriority)
   echo ""
 
   # if the "--installed" flag was passed, don't display remote versions as we didn't fetch data for them.
-  if not params.onlyInstalled:
+  if (not params.onlyInstalled):
     display("Available:", " ", priority = HighPriority)
     for version in remoteVersions:
       if not (version in localVersions):
--- a/src/choosenim.nim
+++ b/src/choosenim.nim
@@ -6,7 +6,7 @@ import nimblepkg/[cli, tools, version]
 import nimblepkg/common as nimbleCommon
 from nimblepkg/packageinfo import getNameVersion
 
-import choosenim/[download, builder, switcher, common, cliparams]
+import choosenim/[download, builder, switcher, common, cliparams, versions]
 import choosenim/[utils, channel, telemetry]
 
 proc parseVersion(versionStr: string): Version =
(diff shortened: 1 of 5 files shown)


src/choosenim/download.nim
```diff
@@ -312,6 +313,16 @@ proc retrieveUrl*(url: string): string =
     var client = newHttpClient(proxy = getProxy())
     return client.getContent(url)
 
+proc getOfficialReleases*(params: CliParams): seq[string] =
+  let rawContents = retrieveUrl(githubTagReleasesUrl)
```
dom96: I don't think this will return the correct versions, it might be better to just hardcode the lower-bound version that we support as nim-lang.org will likely 404 on some of the earliest versions.

src/choosenim.nim
```diff
@@ -196,6 +176,60 @@ proc show(params: CliParams) =
       else:
         display("", ver, priority = HighPriority)
 
+proc versions(params: CliParams) =
+  let currentChannel = getCurrentChannel(params)
+  let currentVersion = getCurrentVersion(params)
+
+  let specialVersions: seq[string] = getSpecialVersions(params)
```
dom96: Nit: Nim has type inference

src/choosenim.nim
```diff
@@ -196,6 +176,60 @@ proc show(params: CliParams) =
...
+  let latestVersion: string =
+    if params.onlyInstalled: ""
+    else: getLatestVersion(params)
+  let remoteVersions: seq[string] =
+    if params.onlyInstalled: @[]
+    else: getAvailableVersions(params)
+
+  #[ Display version information,now that it has been collected ]#
+  proc isActiveTag(params: CliParams, version: string): string =
+    let tag = if version == currentVersion: "*"
```
dom96: Nit: This style is preferable:

```nim
let tag =
  if version == currentVersion: "*"
  else: " "
```

Same for below.

src/choosenim.nim
```diff
@@ -196,6 +176,60 @@ proc show(params: CliParams) =
...
+              else: " "
+    return tag
+  proc isLatestTag(params: CliParams, version: string): string =
+    let tag = if isLatestVersion(params, version): " (latest)"
+            else: ""
+    return tag
+  proc canUpdateTag(params: CliParams, channel: string): string =
```
dom96: Nit: needs empty line between procs.

src/choosenim/versions.nim
```diff
@@ -0,0 +1,55 @@
+
+import os, strutils, algorithm, sequtils
+
+import nimblepkg/version
```
dom96: Please use this module here.

You're reimplementing a lot of its functionality. Also I would prefer if any functionality that you do need works on the ``Version`` type instead of ``string``.

src/choosenim/download.nim
samdmarshall: This returns a listing of all tagged releases on GitHub, without some insight into how the `https://nim-lang.com/channel/...` interface works, I don't know how else to fetch a list of valid releases. I tried a few of the listed versions that got returned from this API and it seemed to work fine. 

if you want to impose a lower-bounds cut-off then that i'm cool with that, again just need some clarity as to how i'm supposed to fetch active/valid version numbers

src/choosenim/versions.nim
samdmarshall: so the action item here, is to use `Version` over `string`, is that correct?

if so, I can make those changes; but I would like to be working against a standardized interface of version numbers. for example, is `"#v0.19.4" == "v0.19.4" == "#0.19.4" == "0.19.4"`??

src/choosenim.nim
samdmarshall: yep, I think that was a left-over from originally when it was just `let specialVersions: seq[string] = @[]`

src/choosenim/download.nim
dom96: Feel free to use the tags, but exclude any version lower than 0.16.0 or something (you can test which versions stop working if you feel up for it).

src/choosenim/versions.nim
dom96: > for example, is "#v0.19.4" == "v0.19.4" == "#0.19.4" == "0.19.4"??

Almost, but not quite. ``#v0.19.4`` is the same as ``0.19.4`` but not semantically, the former means "grab branch/tag `v0.19.4` from github", the latter means "grab version 0.19.4 from nim-lang.org". Internally in Nimble `#foobar` is a "Special" version and is treated differently. You shouldn't really be normalizing these.

samdmarshall: for the record, I think that mechanical difference is awful. that seems it could lead to problems where the language version returned by `vX.Y.Z` is different from that returned by `X.Y.Z` in content based on the source. shouldn't there be one holistic list somewhere?

samdmarshall:
@dom96 okay so I have corrected all the other bits of the PR, so my only remaining issue is how/where to fetch the list of available versions that can be installed. As mentioned, I couldn't work out how to interface with `https://nim-lang.org/channels/etc`, or otherwise see how it could return what available versions it has; so I went with grabbing the GitHub tagged versions. As you mentioned, this could be inconsistent, and additionally the `Version` type defined in `nimblepkg/version` throws when it tries to parse a version number string that is prefixed with `'v'`. So it seems like I should not be using the tag list from GitHub for this, unless part of the `nimblepkg` code gets changed to also handle these version numbers. If we want to include both, I don't mind making it list sources as well; however I don't know how to resolve the issue of where I ***should*** be getting the list from.

Also: is there a way for "special" versions to give more detailed information? eg: I had a `#devel` version installed locally that was from between `0.18.1` and `0.18.2` but it would only tell me `#devel` instead of the git sha. is that a to-do or something that can be done locally?

src/choosenim/versions.nim
dom96: This is just a side-effect of how git works. The purpose of allowing these is to make it possible to install any commit hash.

dom96:
> ... however I don't know how to resolve the issue of where I should be getting the list from.

You should be using the GitHub tags. Nimble does the same thing, you can use this function: https://github.com/nim-lang/nimble/blob/master/src/nimblepkg/download.nim#L105.

> is there a way for "special" versions to give more detailed information? eg: I had a #devel version installed locally that was from between 0.18.1 and 0.18.2 but it would only tell me #devel instead of the git sha. is that a to-do or something that can be done locally?

Don't think so. The `#devel` is just downloaded from GitHub instead of git cloned IIRC so there is no commit info. We would need to query this during download and store it somewhere I guess.

-----

Btw can you also show what the `version` invokation looks like?

samdmarshall:
Oh! i totally forgot about that; here is a screenshot; it displays the current channel (if there is one selected), and lists the selected version with the `*` next to the number. 

Unfiltered:
image

Filtered:
image

dom96:
Nice, looks awesome. If you want to make this even better it would be cool if it showed where each channel lies (there is only two ``stable`` and ``devel``). :)

dom96:
Anyway, I'll just merge this now. Thank you!

PR #117, opened 2019-05-03
Fix eraseLine exception in msys

genotrance:
Choosenim fails on Travis for devel on Windows.

### DIFF
--- a/src/choosenim/download.nim
+++ b/src/choosenim/download.nim
@@ -25,7 +25,7 @@ proc showIndeterminateBar(progress, speed: BiggestInt, lastPos: var int) =
   try:
     eraseLine()
   except OSError:
-    discard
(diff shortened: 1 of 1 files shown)


genotrance:
The exact same commit ran successfully on my Travis instance. Gosh.

https://travis-ci.org/genotrance/choosenim/builds/527930381

dom96:
Shouldn't you fix eraseLine instead? Why would it raise that OSError?

genotrance:
I'll debug that in Nim directly. This fix is needed regardless since eraseLine will fail in a nontty.

dom96:
okay, might be a good idea to write a newline in case this fails then. Otherwise you'll get a super long line with thousands of progress bars, better to have one line each IMO.

PR #132, opened 2019-08-08
Update MinGW link to point to the correct version.

### LINKED ISSUE #129: choosenim installs old version of mingw causing nim cpp to be broken on windows
Problem:
on Windows, (I am using windows 10), choosenim installs Mingw 4.9.1, but the latest devel recommends Mingw 6.3.0

Using the old version of Mingw results in `nim cpp` producing broken exe's.
Nim will compile your program, but running it produces no output. Even a simple hello world program does not work.

Manually downloading and installing Mingw 6.3.0 from the nim website, and forcing choosenim to use that version fixes the problem.

Steps:
1. Clear out out the `C:\User\<username>\.nimble` and `C:\User\<username>\.choosenim` directories first
2. Download the latest windows release (v 0.4.0) from the github release page
3. Extract and run `runme.bat`
4. Open Powershell 
5. run `choosenim update "#head"`
6. Run `nim cpp -r hello_world.nim` -> this fails

Steps to force Mingw 6.3.0
1. remove `g++.exe` and `gcc.exe` from `C:\User\<username>\.nimble`
(I used Get-Command to determine which g++ was being used. This is the powershell equivalent of the Unix `which` command)
2. Download Mingw 6.3.0 from https://nim-lang.org/install_windows.html
3. Extract the zip file
4. Open Powershell
4. Add the Mingw directory to your windows path.
In Powershell I did the following to temporarily add to my path: `$env:Path += "C:\Users\ray\Downloads\mingw64-6.3.0\mingw64\bin"`
5. run `choosenim update "#head"`
6. Run `nim cpp -r hello_world.nim` -> this now succeeds

rayman22201:
closes #129

### DIFF
--- a/src/choosenim/download.nim
+++ b/src/choosenim/download.nim
@@ -14,7 +14,7 @@ const
   csourcesUrl = "https://github.com/nim-lang/csources/archive/master.tar.gz"
 
 const # Windows-only
-  mingwUrl = "http://nim-lang.org/download/mingw32.tar.gz"
+  mingwUrl = "http://nim-lang.org/download/mingw32-6.3.0.7z"
   dllsUrl = "http://nim-lang.org/download/dlls.tar.gz"
   winBinaryZipUrl = "http://nim-lang.org/download/nim-$1_x$2.zip"
 


dom96:
Sadly choosenim cannot handle extraction of .7z files. I think the solution is to simply update the mingw32.tar.gz file on nim-lang.org

rayman22201:
Bah. I forgot about that. Closing this PR.

PR #138, opened 2019-09-25
dump the needed foreign dependencies for common distros

Xe:
Based on comments and feedback from IRC

### DIFF
--- a/src/choosenim.nim
+++ b/src/choosenim.nim
@@ -1,12 +1,12 @@
 # Copyright (C) Dominik Picheta. All rights reserved.
 # BSD-3-Clause License. Look at license.txt for more info.
-import os, strutils, algorithm
+import distros, os, strutils, algorithm
 
 import nimblepkg/[cli, tools, version]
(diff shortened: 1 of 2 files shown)


src/choosenim.nim
```diff
@@ -40,6 +40,9 @@ proc chooseVersion(version: string, params: CliParams) =
               Warning, HighPriority)
       display("Hint:", "Install clang or gcc using your favourite package manager.",
               Warning, HighPriority)
+      display("Hint:", "The following commands might help:", Warning, LowPriority)
+
+      echoForeignDeps()
```
PMunch: Won't this echo no matter the verbosity level, while the above command will only show the explanation for Warning and LowPriority?

dom96: Just don't use the `foreignDep`, problem solved, no?

genotrance:
Any update on this PR?

dom96:
Please reopen if you get a chance to correct the verbosity problem.

PR #143, opened 2019-10-25
Use nimarchive, linux binary builds, mingw 7z and windows CI

genotrance:


.travis.yml
```diff
@@ -1,31 +1,57 @@
...
+    if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+      # Work around https://github.com/nim-lang/Nim/issues/12337 fixed in 1.0+
+      ulimit -n 8192
+    fi
 
 script:
   - nimble test
   - yes | choosenim stable # Workaround tester overwriting our Nimble install.
-  - nimble build -d:release
+  - nimble build -d:danger
```
dom96: I don't think this is worth the danger.

src/choosenim.nim
```diff
@@ -16,7 +16,10 @@ proc installVersion(version: Version, params: CliParams) =
   let extractDir = params.getInstallationDir(version)
   # Make sure no stale files from previous installation exist.
   removeDir(extractDir)
-  extract(path, extractDir)
+  if version.isSpecial:
+    extract(path, extractDir)
+  else:
+    extract(path, extractDir.parentDir())
```
dom96: This seems like you're working around a problem in ``extract``. IIRC I handled this exact scenario in my ``untar`` package, shouldn't it also be handled in nimarchive?

src/choosenim/download.nim
```diff
@@ -9,14 +9,13 @@ import cliparams, common, switcher, telemetry, utils
...
-    getArchiveFormat()
+  websiteUrl = "http://nim-lang.org/download/nim-$1.tar.xz"
   csourcesUrl = "https://github.com/nim-lang/csources/archive/master.tar.gz"
+  binaryUrl = "http://nim-lang.org/download/nim-$1$2_x$3" & getBinArchiveFormat()
 
 const # Windows-only
-  mingwUrl = "http://nim-lang.org/download/mingw32.tar.gz"
-  dllsUrl = "http://nim-lang.org/download/dlls.tar.gz"
-  winBinaryZipUrl = "http://nim-lang.org/download/nim-$1_x$2.zip"
+  mingwUrl = "http://nim-lang.org/download/mingw32-6.3.0.7z"
```
dom96: Shouldn't we hardcode a file path that doesn't refer to a version so that it can be updated without needing a new version of choosenim?

src/choosenim/download.nim
```diff
@@ -226,17 +225,16 @@ proc downloadImpl(version: Version, params: CliParams): string =
...
-      if findExe("powershell") != "":
-        let winUrl = winBinaryZipUrl % [$version, $arch]
-        if not needsDownload(params, winUrl, outputPath): return outputPath
-        try:
-          downloadFile(winUrl, outputPath, params)
-          return outputPath
-        except HttpRequestError:
-          display("Info:", "Binary build unavailable, building from source",
-                  priority = HighPriority)
+    when defined(Windows) or defined(linux):
```
dom96: Nit: Please add comment above here, like:

```
# Use binary builds for Linux and Windows.
```

src/choosenim.nim
dom96: Happy for you to add a comment above this if statement with a link to an issue in nimarchive and resolve this in the future.

genotrance: The way you handle this is to extract to a temporary directory and then copy the subdir contents back into the toolchains directory. That's a double copy that takes  a bunch of time, especially for xz files which are huge once extracted.

Most extraction tools simply extract the contents as is and it is up to the consumer to do this. I made a simple change based on whether it is a github archive which we know doesn't have the subdir or any of our xz files which do.

src/choosenim/download.nim
genotrance: I'm cool with that but need a URL I should put in there that exists.

.travis.yml
genotrance: Done.

src/choosenim/download.nim
genotrance: Done

src/choosenim.nim
dom96: The problem here is that we cannot know for sure that our archives will or won't have the subdir, the more I think about this the more this solution seems unreliable to me. You are also working around this by extracting to the parent dir? That seems really bad, are you just assuming that the directory will happen to match what choosenim expects?

src/choosenim/download.nim
dom96: There is apparently already a `http://nim-lang.org/download/mingw32.7z`, it's bigger than the 6.3.0 so I'm not sure what version it is, but let's use that if it works. Otherwise I can overwrite it with 6.3.0.

CC @Araq

genotrance: The mingw32.7z archive is still 4.8.1, older than the 6.3.0 that we advertise on the website.

Can you please symlink mingw32.7z and mingw64.7z to the 6.3.0 versions for now? Araq is open to updating to 8.1.0 and once that is done, we can update the symlinks.

src/choosenim.nim
genotrance: `unzip -j` is almost useless, it skips ALL directory structures, throwing everything into the destination leading to overwrites. That's why I felt it should be handled in the consumer. I also don't quite agree with the fact that our archives will have different structures - the CI is consistent. 

Regardless, `skipOuterDirs` as implemented does work well what with our 3 mingw archives (gz, 7z, versioned 7z) using different subdir names, plus 32/64 bit naming, so I agree with that assumption being fragile.

I'll add the feature to nimarchive and update this PR.

src/choosenim/download.nim
dom96: Done.

genotrance: Awesome. Do we also want to update the tar.gz to 6.3.0 as well so that older users of choosenim also benefit?

https://github.com/dom96/choosenim/issues/129

src/choosenim.nim
genotrance: Fixed

src/choosenim/download.nim
genotrance: Fixed.

PR #156, opened 2020-01-10
Multiple fixes

### LINKED ISSUE #147: OS error: Permission denied at version `0.5`
`Spawning of process failed. (Error was: Additional info: "Could not find command: \'~/.choosenim/toolchains/nim-1.0.4/bin/nimble\'. OS error: Permission denied`

### LINKED ISSUE #151: CI does not name binaries correctly
The CI should name the binaries in the same way as is done for all previous releases.

CC @genotrance

genotrance:
Fix #147 - workaround rather since we don't have root cause yet
Fix #151 - correct binary names in CI
Add testament to list of proxies

### DIFF
--- a/src/choosenim/builder.nim
+++ b/src/choosenim/builder.nim
@@ -73,6 +73,19 @@ proc buildTools() =
     else:
       doCmdRaw("./koch tools -d:release")
 
+# Workaround for #147
+when defined(posix):
+  import posix
+
+  proc setPermissions() =
+    ## Assumes that CWD contains the compiler
+    let binDir = getCurrentDir() / "bin"
+    for kind, path in walkDir(binDir):
+      if kind == pcFile:
+        # Set to 755 = rwxrw-rw- in octal
+        # = 493 in decimal
+        discard chmod(path, 493)
+
 proc build*(extractDir: string, version: Version, params: CliParams) =
   # Report telemetry.
   report(initEvent(BuildEvent), params)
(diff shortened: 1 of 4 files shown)


src/choosenim/builder.nim
```diff
@@ -73,6 +73,19 @@ proc buildTools() =
...
+  import posix
+
+  proc setPermissions() =
+    ## Assumes that CWD contains the compiler
+    let binDir = getCurrentDir() / "bin"
+    for kind, path in walkDir(binDir):
+      if kind == pcFile:
+        # Set to 755 = rwxrw-rw- in octal
+        # = 493 in decimal
+        discard chmod(path, 493)
```
dom96: You shouldn't need posix for this, can you just use the `os` module?

genotrance: Fixed.

src/choosenim/builder.nim
```diff
@@ -73,6 +73,19 @@ proc buildTools() =
...
+    ## Assumes that CWD contains the compiler
+    let binDir = getCurrentDir() / "bin"
+    for kind, path in walkDir(binDir):
+      if kind == pcFile:
+        setFilePermissions(path,
+                           {fpUserRead, fpUserWrite, fpUserExec,
+                            fpGroupRead, fpGroupExec,
+                            fpOthersRead, fpOthersExec}
+        )
+
```
dom96: Another thing that would be nice would be a `debug` output to say that these file permission changes are taking place. But I'll leave that to another PR if you agree that it's a good idea.

genotrance: To do that, we will have to first check the extracted permissions. Do you want to add that additional code?

dom96: Huh? Just show the message always, even if the permissions aren't being changed. It'll be hidden anyway most of the time.

genotrance: Done in PR #158.

PR #158, opened 2020-01-12
Fix #98 - remove c_code after build

### LINKED ISSUE #98: Clean up after building a version to save space
Talked to someone with not a whole lot of space left on their device that wanted to give Nim a try today. Checked how much `choosenim` was using on my system and it showed a pretty steep 4.1G of used space. I've got about 9 versions installed, but 0.19.0 alone is almost half a gig large. The main culprit for each version appears to be the c_code folder. The downloads folder is another half gig combined. It would be nice if choosenim could clean up these files that are no longer used for anything so that it didn't hog as much space.

genotrance:


src/choosenim/builder.nim
```diff
@@ -110,6 +110,11 @@ proc build*(extractDir: string, version: Version, params: CliParams) =
     buildTools()
     when defined(posix):
       setPermissions() # workaround for #147
+    # Delete c_code
+    try:
+      removeDir(extractDir / "c_code")
+    except OSError:
+      discard
```
dom96: Move this to the `finally` instead below.

src/choosenim/builder.nim
```diff
@@ -85,6 +85,7 @@ when defined(posix):
                             fpGroupRead, fpGroupExec,
                             fpOthersRead, fpOthersExec}
         )
+        display("Info", "Setting execution permissions", Message, LowPriority)
```
dom96: please include the path in this message

src/choosenim/builder.nim
```diff
@@ -120,6 +121,12 @@ proc build*(extractDir: string, version: Version, params: CliParams) =
     raise newError
   finally:
     if success:
+      # Delete c_code
+      try:
+        removeDir(extractDir / "c_code")
+      except OSError:
+        discard
```
dom96: Also, add a warning here as done in the `try` that's also removing stuff below please.

src/choosenim/builder.nim
genotrance: Done

src/choosenim/builder.nim
genotrance: Done

src/choosenim/builder.nim
genotrance: Done

PR #160, opened 2020-01-12
Post debug builds to releases, bump required versions

genotrance:


.travis.yml
```diff
@@ -24,13 +24,14 @@ script:
   - nim c src/choosenim
   - nimble test
```
dom96: we might need to ensure that `nimble test` doesn't recompile in release mode via `nimble build` or something.

genotrance: Tester builds with `nimble build` which does not create a release build. We later run `nimble build -d:release` which does. CI output shows the different file sizes.

https://travis-ci.org/genotrance/choosenim/jobs/636552336#L2824

PR #164, opened 2020-01-14
Fix #28 - add ~/.nimble/bin to PATH on Windows

### LINKED ISSUE #28: Ask the user whether they want Nim added to PATH
Also perform the modification of PATH automatically on all platforms.

genotrance:
image

### DIFF
--- a/src/choosenimpkg/switcher.nim
+++ b/src/choosenimpkg/switcher.nim
@@ -120,7 +120,7 @@ when defined(windows):
               nimbleDesiredPath & " is not in your PATH environment variable.\n" &
               "            Should it be added permanently?"):
       addToPathEnv(nimbleDesiredPath)
-      display("NOTE:", "PATH changes will only take effect in new sessions.",
+      display("Note:", "PATH changes will only take effect in new sessions.",
               priority = HighPriority)
 
 proc isNimbleBinInPath*(params: CliParams): bool =
@@ -183,7 +183,7 @@ proc writeProxy(bin: string, params: CliParams) =
   try:
     writeFile(proxyPath, proxyExe)
   except IOError:
-    display("Warning:", "component '$1' in use, write failed" % bin, Warning,
+    display("Warning:", "component '$1' possibly in use, write failed" % bin, Warning,
             priority = HighPriority)
     return
 
@@ -199,7 +199,7 @@ proc writeProxy(bin: string, params: CliParams) =
       when defined(windows):
         "Binary '$1' isn't in your PATH" % bin
       else:
-        "Binary '$1' isn't in your PATH. Add '$2' to your PATH." %
+        "Binary '$1' isn't in your PATH. Ensure that '$2' is in your PATH." %
           [bin, params.getBinDir()]
     display("Hint:", msg, Warning, HighPriority)
   elif fromPATH != "" and fromPATH != proxyPath:
--- a/src/choosenimpkg/switcher.nim
+++ b/src/choosenimpkg/switcher.nim
@@ -5,6 +5,9 @@ from nimblepkg/packageinfo import getNameVersion
 
 import cliparams, common
 
+when defined(windows):
+  import env
+
 proc compileProxyexe() =
   var cmd = "nim c"
   when defined(release):
@@ -89,68 +92,6 @@ proc getNimbleVersion(toolchainPath: string): Version =
   else:
     display("Warning:", "Could not find toolchain's Nimble version.",
             Warning, MediumPriority)
-
-when defined(windows):
-  # From finish.nim in nim-lang/Nim/tools
-  import registry
-
-  proc tryGetUnicodeValue(path, key: string, handle: HKEY): string =
-    # Get a unicode value from the registry or ""
-    try:
-      result = getUnicodeValue(path, key, handle)
-    except:
-      result = ""
-
-  proc addToPathEnv(e: string) =
-    # Append e to user PATH to registry
-    var p = tryGetUnicodeValue(r"Environment", "Path", HKEY_CURRENT_USER)
-    let x = if e.contains(Whitespace): "\"" & e & "\"" else: e
-    if p.len > 0:
-      if p[^1] != PathSep:
-        p.add PathSep
-      p.add x
-    else:
-      p = x
-    setUnicodeValue(r"Environment", "Path", p, HKEY_CURRENT_USER)
-
-  proc setNimbleBinPath*(params: CliParams) =
-    # Ask the user and add nimble bin to PATH
-    let nimbleDesiredPath = params.getBinDir()
-    if prompt(params.nimbleOptions.forcePrompts,
-              nimbleDesiredPath & " is not in your PATH environment variable.\n" &
-              "            Should it be added permanently?"):
-      addToPathEnv(nimbleDesiredPath)
-      display("Note:", "PATH changes will only take effect in new sessions.",
-              priority = HighPriority)
-
-proc isNimbleBinInPath*(params: CliParams): bool =
-  # This proc searches the $PATH variable for the nimble bin directory,
-  # typically ~/.nimble/bin
-  result = false
-  let nimbleDesiredPath = params.getBinDir()
-  when defined(windows):
-    let p = tryGetUnicodeValue(r"Environment", "Path",
-      HKEY_CURRENT_USER) & PathSep & tryGetUnicodeValue(
-      r"System\CurrentControlSet\Control\Session Manager\Environment", "Path",
-      HKEY_LOCAL_MACHINE)
-  else:
-    let p = getEnv("PATH")
-  for x in p.split(PathSep):
-    if x.len == 0: continue
-    let y =
-      try:
-        expandFilename(
-          if x[0] == '"' and x[^1] == '"':
-            substr(x, 1, x.len-2)
-          else: x
-        )
-      except OSError as e:
-        if e.errorCode == 0: x
-        else: ""
-      except: ""
-    if y.cmpIgnoreCase(nimbleDesiredPath) == 0:
-      result = true
-      break
 
 proc writeProxy(bin: string, params: CliParams) =
   # Create the ~/.nimble/bin dir in case it doesn't exist.
--- a/src/choosenimpkg/switcher.nim
+++ b/src/choosenimpkg/switcher.nim
@@ -93,6 +93,68 @@ proc getNimbleVersion(toolchainPath: string): Version =
     display("Warning:", "Could not find toolchain's Nimble version.",
             Warning, MediumPriority)
 
+when defined(windows):
+  # From finish.nim in nim-lang/Nim/tools
+  import registry
+
+  proc tryGetUnicodeValue(path, key: string, handle: HKEY): string =
+    # Get a unicode value from the registry or ""
+    try:
+      result = getUnicodeValue(path, key, handle)
+    except:
+      result = ""
+
+  proc addToPathEnv(e: string) =
+    # Append e to user PATH to registry
+    var p = tryGetUnicodeValue(r"Environment", "Path", HKEY_CURRENT_USER)
+    let x = if e.contains(Whitespace): "\"" & e & "\"" else: e
+    if p.len > 0:
+      if p[^1] != PathSep:
+        p.add PathSep
+      p.add x
+    else:
+      p = x
+    setUnicodeValue(r"Environment", "Path", p, HKEY_CURRENT_USER)
+
+  proc setNimbleBinPath*(params: CliParams) =
+    # Ask the user and add nimble bin to PATH
+    let nimbleDesiredPath = params.getBinDir()
+    if prompt(params.nimbleOptions.forcePrompts,
+              nimbleDesiredPath & " is not in your PATH environment variable.\n" &
+              "            Should it be added permanently?"):
+      addToPathEnv(nimbleDesiredPath)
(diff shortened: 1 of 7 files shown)


src/choosenimpkg/switcher.nim
```diff
@@ -84,6 +84,37 @@ proc getNimbleVersion(toolchainPath: string): Version =
...
+                  Warning, HighPriority)
+
+  proc setNimbleBinPath*(params: CliParams) =
+    let nimbleDesiredPath = params.getBinDir()
+    if prompt(params.nimbleOptions.forcePrompts,
+              nimbleDesiredPath & " is not in your PATH environment variable.\n" &
+              "            Should it be added permanently?"):
+      addToPathEnv(nimbleDesiredPath)
+
+proc isNimbleBinInPath*(params: CliParams): bool =
```
dom96: I almost suggested `findExe` here but then I noticed that you are checking for the nimble bin path very specifically: we don't want to have duplicates added. Can you add a doc comment to this function to describe what it does?

src/choosenimpkg/switcher.nim
```diff
@@ -84,6 +84,37 @@ proc getNimbleVersion(toolchainPath: string): Version =
     display("Warning:", "Could not find toolchain's Nimble version.",
             Warning, MediumPriority)
 
+when defined(windows):
+  proc addToPathEnv(e: string) =
+    let exitCode = execCmdEx("cmd /c setx PATH " & e.quoteShell).exitCode
```
dom96: Won't this overwrite the PATH with just the filepath to Nimble's bin dir? Don't you need `"%PATH%;"` in there?

src/choosenimpkg/switcher.nim
```diff
@@ -84,6 +84,37 @@ proc getNimbleVersion(toolchainPath: string): Version =
...
+    if exitCode != 0:
+      display("Warning:", "Failed to add to PATH")
+    else:
+      display("Note:", "PATH changes will only take effect in new sessions.",
+                  Warning, HighPriority)
```
dom96: Can't you also use `set` to set it in the current session? If you do that then you'll be able to double check that the PATH was updated by using `isNimbleBinInPath`

src/choosenimpkg/switcher.nim
```diff
@@ -84,6 +84,37 @@ proc getNimbleVersion(toolchainPath: string): Version =
...
+  result = false
+  let nimbleDesiredPath = params.getBinDir()
+  let p = getEnv("PATH")
+  for x in p.split(PathSep):
+    if x.len == 0: continue
+    let y = try: expandFilename(if x[0] == '"' and x[^1] == '"':
+                                  substr(x, 1, x.len-2) else: x)
+            except OSError as e:
+              if e.errorCode == 0: x else: ""
+            except: ""
```
dom96: Please reformat like this:

```
            let y =
              try:
                expandFilename(
                  if x[0] == '"' and x[^1] == '"':
                    substr(x, 1, x.len-2)
                  else: x
                )
              except OSError as e:
                if e.errorCode == 0: x
                else: ""
              except: ""

Comment (dom96) [reply src/choosenimpkg/switcher.nim]: or maybe even get rid of so much nesting.

File: src/choosenimpkg/switcher.nim (hunk 2 above)
Comment (genotrance) [reply src/choosenimpkg/switcher.nim]: Thanks for catching this - it's completely wrong and even if I added everything, it is wrong. I'll fix this post 0.5.1.

File: src/choosenimpkg/switcher.nim (hunk 3 above)
Comment (genotrance) [reply src/choosenimpkg/switcher.nim]: Since I'm prompting the user to add, I'm not adding it to the main path. Else, every time the user will be asked if he opts out for any reason.

File: src/choosenimpkg/switcher.nim (hunk 4 above)
Comment (genotrance) [reply src/choosenimpkg/switcher.nim]: Changing formatting but keeping it the same since I got it from finish.nim.

File: src/choosenimpkg/switcher.nim (hunk 1 above)
Comment (genotrance) [reply src/choosenimpkg/switcher.nim]: Done

File: src/choosenimpkg/switcher.nim (hunk 2 above)
Comment (genotrance) [reply src/choosenimpkg/switcher.nim]: This has been updated to use finish's algorithm.

**Comment by genotrance** [comment]
This is now ready to go.

File: src/choosenimpkg/switcher.nim [hunk 5]
```diff
@@ -87,6 +87,68 @@ proc getNimbleVersion(toolchainPath: string): Version =
...
+
+  proc setNimbleBinPath*(params: CliParams) =
+    # Ask the user and add nimble bin to PATH
+    let nimbleDesiredPath = params.getBinDir()
+    if prompt(params.nimbleOptions.forcePrompts,
+              nimbleDesiredPath & " is not in your PATH environment variable.\n" &
+              "            Should it be added permanently?"):
+      addToPathEnv(nimbleDesiredPath)
+      display("NOTE:", "PATH changes will only take effect in new sessions.",
+              priority = HighPriority)
```
dom96: This should go in a separate module.

src/choosenimpkg/switcher.nim
```diff
@@ -87,6 +87,68 @@ proc getNimbleVersion(toolchainPath: string): Version =
...
+    setUnicodeValue(r"Environment", "Path", p, HKEY_CURRENT_USER)
+
+  proc setNimbleBinPath*(params: CliParams) =
+    # Ask the user and add nimble bin to PATH
+    let nimbleDesiredPath = params.getBinDir()
+    if prompt(params.nimbleOptions.forcePrompts,
+              nimbleDesiredPath & " is not in your PATH environment variable.\n" &
+              "            Should it be added permanently?"):
+      addToPathEnv(nimbleDesiredPath)
+      display("NOTE:", "PATH changes will only take effect in new sessions.",
```
dom96: Lowercase. It's 2020, not 2005 :)
```suggestion
      display("Note:", "PATH changes will only take effect in new sessions.",
```

src/choosenimpkg/switcher.nim
```diff
@@ -87,6 +87,68 @@ proc getNimbleVersion(toolchainPath: string): Version =
...
+proc isNimbleBinInPath*(params: CliParams): bool =
+  # This proc searches the $PATH variable for the nimble bin directory,
+  # typically ~/.nimble/bin
+  result = false
+  let nimbleDesiredPath = params.getBinDir()
+  when defined(windows):
+    let p = tryGetUnicodeValue(r"Environment", "Path",
+      HKEY_CURRENT_USER) & PathSep & tryGetUnicodeValue(
+      r"System\CurrentControlSet\Control\Session Manager\Environment", "Path",
+      HKEY_LOCAL_MACHINE)
```
dom96: What is this and why is it necessary? `getEnv"PATH"` should be enough.

src/choosenimpkg/switcher.nim
```diff
@@ -87,6 +87,68 @@ proc getNimbleVersion(toolchainPath: string): Version =
...
+  else:
+    let p = getEnv("PATH")
+  for x in p.split(PathSep):
+    if x.len == 0: continue
+    let y =
```
dom96: Nit ``y`` and ``x`` are not descriptive enough, consider using more verbose names.

src/choosenimpkg/switcher.nim
```diff
@@ -87,6 +87,68 @@ proc getNimbleVersion(toolchainPath: string): Version =
...
+      try:
+        expandFilename(
+          if x[0] == '"' and x[^1] == '"':
+            substr(x, 1, x.len-2)
+          else: x
+        )
+      except OSError as e:
+        if e.errorCode == 0: x
+        else: ""
+      except: ""
```
dom96: This can be significantly simplified:

```nim
let stripped = strip(x, chars = {"\""})
let expandedPath =
  try:
    expandFilename(stripped)
  except:
    "" # IMO if OSError is raised with an errorCode of 0 then that's a stdlib bug, and you shouldn't hack around it
```

src/choosenimpkg/switcher.nim
```diff
@@ -115,7 +177,13 @@ proc writeProxy(bin: string, params: CliParams) =
   # Don't write the file again if it already exists.
   if fileExists(proxyPath) and readFile(proxyPath) == proxyExe: return
 
-  writeFile(proxyPath, proxyExe)
+  try:
+    writeFile(proxyPath, proxyExe)
+  except IOError:
+    display("Warning:", "component '$1' in use, write failed" % bin, Warning,
```
dom96:
```suggestion
    display("Warning:", "component '$1' possibly in use, write failed" % bin, Warning,
```

src/choosenimpkg/switcher.nim
```diff
@@ -124,8 +192,13 @@ proc writeProxy(bin: string, params: CliParams) =
...
   if fromPATH == "" and not params.firstInstall:
-      display("Hint:", "Binary '$1' isn't in your PATH. Add '$2' to your PATH." %
-              [bin, params.getBinDir()], Warning, HighPriority)
+    let msg =
+      when defined(windows):
+        "Binary '$1' isn't in your PATH" % bin
+      else:
+        "Binary '$1' isn't in your PATH. Add '$2' to your PATH." %
+          [bin, params.getBinDir()]
+    display("Hint:", msg, Warning, HighPriority)
```
dom96: choosenim could fail to add to the PATH. This change makes the code more complex and reduces the amount of info the user gets.

You should change the text to "Binary '$1' isn't in your PATH. Ensure that '$2' is in your PATH." instead

src/choosenimpkg/switcher.nim
```diff
@@ -178,6 +251,11 @@ proc switchToPath(filepath: string, params: CliParams): bool =
   for proxy in proxiesToInstall:
     writeProxy(proxy, params)
 
+  when defined(windows):
+    if not isNimbleBinInPath(params):
+      display("Hint:", "Use 'choosenim <version/channel> --firstInstall' to add\n" &
+                  "$1 to your PATH." % params.getBinDir(), Warning, HighPriority)
```
dom96: Cool, thanks for adding this hint :)

src/choosenimpkg/switcher.nim
genotrance: Do you want a brand new module? I cannot put this in utils since it imports switcher but this code is called by switcher. I could put it in cliparams but it isn't the right place either.

src/choosenimpkg/switcher.nim
genotrance: Done

src/choosenimpkg/switcher.nim
genotrance: I got this code from finish.exe. $PATH could always be modified so I think @Araq got this from the registry to get the original value.

src/choosenimpkg/switcher.nim
genotrance: Again, I got this from finish.nim so didn't diverge to keep consistent.

src/choosenimpkg/switcher.nim
genotrance: If you want to diverge from finish.nim, I can do this.

src/choosenimpkg/switcher.nim
genotrance: Done

src/choosenimpkg/switcher.nim
genotrance: I reduced the output since the hint to use --firstInstall will appear at the end.

Changed the text.

src/choosenimpkg/switcher.nim
dom96: yes please

src/choosenimpkg/switcher.nim
dom96: Yes, please do so.

src/choosenimpkg/switcher.nim
dom96: Please put it in the separate module at least. Do simplify the below code though.

src/choosenimpkg/switcher.nim
dom96: Leave a comment about this then. Otherwise someone eventually will simplify this code.

To be honest, I don't get why you'd want to get the original value. The only way $PATH could have been modified is if the user launches choosenim via ``PATH=$PATH:/blah/bin``. Am I missing something?

src/choosenimpkg/switcher.nim
dom96: Then this hint should be shown here instead of somewhere downstream. Don't you think?

genotrance: Couldn't do that since this proc gets called for each proxy target so the hint would show up multiple times.

dom96: IIRC ``display`` deduplicates messages, so you can call it as many times as you want with the same message, it'll only be shown once.

src/choosenimpkg/switcher.nim
genotrance: Done - added env.nim

src/choosenimpkg/switcher.nim
genotrance: $PATH could be modified by something else earlier in the session as well. Given we are setting this to the registry, we should get from there as well. I've added a comment.

src/choosenimpkg/switcher.nim
genotrance: Done

src/choosenimpkg/switcher.nim
genotrance: Done

PR #167, opened 2020-01-28
Travis regression test

iranika:
How about doing a regression test with travis ci?
For example, `choosenim update self`.
```.travis.yml
script:
  ~~~
  - sed -i -e s@chooseNimVersion.*@chooseNimVersion\*\ =\ \"0.4.0\"@ src/choosenimpkg/common.nim
  - cat src/choosenimpkg/common.nim
  - choosenim stable
  - nimble build -d:release
  - which choosenim
  - cp -pr bin/choosenim${EXT} ~/.nimble/bin/
  - choosenim --version
  - choosenim update self --debug
  - choosenim --version
```

The CI test will be longer, but I think quality control will be easier.
~~Now, my .travis.yml is a draft. I want to PullRequest if it gets better.~~
I open Pull Request.
Please comment if you have any opinions or ideas.
thanks! =)


genotrance:
Looks interesting, please clean it up and submit.

iranika:
For now, I have cleaned it.
Usually using `jobs:` might be wise, but I didn't do it this time because it was a big change.
If we want to deal with cases like # 168, we should consider testing only on Windows using `jobs:`.
I need to learn TravisCI a bit more :(

For now, let's postpone `jobs:`.

.travis.yml
```diff
@@ -34,6 +34,31 @@ script:
   - nimble build -d:release
```
genotrance: Why not move directly to artifact here instead of copying later on?

.travis.yml
```diff
@@ -34,6 +34,31 @@ script:
...
   - mv "bin/choosenim${EXT}" "bin/choosenim-${VERSION}_${OSNAME}_${ARCH}${EXT}"
   - ls -l bin
+  - mkdir artifact
+  - cp -pr bin/* artifact/ #Copy release files to artifact dir for deploy.
+  - ls -l artifact
+  - echo "Build test end." 
+  - echo "Regression test start."
+  # switch and show test.
```
genotrance: Regression was already done on line 25.

.travis.yml
```diff
@@ -34,6 +34,31 @@ script:
...
+  - choosenim --version
+  - choosenim 0.20.0 
+  - choosenim show
+  - choosenim stable
+  # test `update self` need rewrite lower version and build. So, run this test last.
```
genotrance: This switching test was already verified in line 25.

.travis.yml
```diff
@@ -34,6 +34,31 @@ script:
...
+  - sed -i -e s@chooseNimVersion.*@chooseNimVersion\*\ =\ \"0.4.0\"@ src/choosenimpkg/common.nim
+  - cat src/choosenimpkg/common.nim
+  - choosenim stable
+  - nimble build -d:release
+  - which choosenim
+  - cp -pr bin/choosenim${EXT} ~/.nimble/bin/
+  - choosenim --version
+  - choosenim update self --debug
+  - choosenim --version
+
```
genotrance: This test can be done in tests/tester.nim instead of this yml file.

.travis.yml
```diff
@@ -34,6 +34,31 @@ script:
...
+after_failure: 
+  - "cat ~/build/iranika/choosenim/src/choosenimpkg/common.nim"
+  - "ls -al"
+  - "pwd"
```
genotrance: Remove?

src/choosenim.nim
```diff
@@ -80,7 +80,7 @@ proc updateSelf(params: CliParams) =
 
   # https://stackoverflow.com/a/9163044/492186
   let tag = "v" & $version
-  let filename = "choosenim-" & $version & "_" & hostOS & "_" & hostCPU
+  let filename = "choosenim-" & $version & "_" & hostOS & "_" & hostCPU.addFileExt(ExeExt)
   let url = "https://github.com/dom96/choosenim/releases/download/$1/$2" % [
```
genotrance: This is from an earlier commit, please rebase.

.travis.yml
iranika: @genotrance Does it mean I should `nimble build -d:release && cp bin/* artifact/` here ?

.travis.yml
iranika: Sorry, I mistake.

.travis.yml
genotrance: No, next step where we are renaming choosenim binary with the right version, osname and arch. Simply change `bin` to `artifacts`. Need to `mkdir artifacts` earlier though.

.travis.yml
iranika: @genotrance  I think it is necessary to rewrite the version of choosenim and build.
Can we do it with tester.nim ?

.travis.yml
iranika: oh, I'm idiot : (
Like this. thanks!
```
  - mkdir artifact
  - mv "bin/choosenim${EXT}" "artifact/choosenim-${VERSION}_${OSNAME}_${ARCH}_debug${EXT}"
  - nimble build -d:release
  - mv "bin/choosenim${EXT}" "artifact/choosenim-${VERSION}_${OSNAME}_${ARCH}${EXT}"
  - ls -l artifact
```

.travis.yml
iranika: OK, Erase it.

.travis.yml
iranika: Yes, I remove. thanks.

.travis.yml
iranika: I want to complete only with tester.nim. It ’s beautiful and simple.
Now, I retried it.
But, `common.chooseNimVersion` is defined in const. So it cannot be changed. Isn't it ?
I don't know how to hack this.
I can only think this one :(
```.travis.yml
  - sed -i -e s@chooseNimVersion.*@chooseNimVersion\*\ =\ \"0.4.0\"@ src/choosenimpkg/common.nim
  - cat src/choosenimpkg/common.nim
  - choosenim stable
  - nimble build -d:release
  - which choosenim
  - cp -pr bin/choosenim${EXT} ~/.nimble/bin/
  - choosenim --version #show version for log
  - choosenim update self --debug
  - choosenim --version #show version for log
```

genotrance: All the same steps above can just be done in pure Nim in tester.nim. So you'd edit the version, make a build and kick off update and verify that it got updated. In fact, you can do some real logic and verify that the version changed as expected. In the above script, no real check is happening. It is just being printed so the CI is only checking that it isn't crashing.

iranika: Sorry, I finally understand. You are right.
Then I don't need to edit this yml. I should be edit tester.nim.
I will later consider starting a new pull request or editing tester.nim with this pull request.
Thanks @genotrance

iranika:
I opened #169 . This Pull Request close.
Thank you for all your help. : )

PR #168, opened 2020-01-30
Fix #128 - detect arch on Windows for mingw

### LINKED ISSUE #128: proposal: on 64-bit Windows, download 64-bit MinGW & Nim
As far as I understand, choosenim currently [just downloads a 32-bit version of MinGW on Windows](https://github.com/dom96/choosenim/blob/121133ba06b0954e058d38743a18fe1ad1f817f2/src/choosenim.nim#L36) (if no gcc is found in PATH), and then, as a consequence, a 32-bit version of Nim. Is there any interest in allowing choosenim to download a 64-bit version of MinGW (and Nim)?

If yes, should there be a flag for choosing the variant, or should the 64-bit toolchain be the default one on 64-bit Windows ([autodetected](https://blogs.msdn.microsoft.com/david.wang/2006/03/27/howto-detect-process-bitness/))?

genotrance:
Installs mingw32 on 32-bit Windows and mingw64 on 64-bit Windows. This will also work correctly if someone builds a 32-bit version of choosenim and runs it on 64-bit Windows.

@dom96 @Araq

### DIFF
--- a/src/choosenimpkg/cliparams.nim
+++ b/src/choosenimpkg/cliparams.nim
@@ -113,6 +113,8 @@ proc getCpuArch*(): int =
   if cpuArch != 0:
     return cpuArch
 
+  var
+    failMsg = ""
   when defined(windows):
     let
       archEnv = getEnv("PROCESSOR_ARCHITECTURE")
@@ -125,6 +127,9 @@ proc getCpuArch*(): int =
       result = 64
     elif "86" in archEnv:
       result = 32
+    else:
+      failMsg = "PROCESSOR_ARCHITECTURE = " & archEnv &
+                ", PROCESSOR_ARCHITEW6432 = " & arch6432Env
   else:
     let
       uname = findExe("uname")
@@ -136,9 +141,15 @@ proc getCpuArch*(): int =
           result = 64
         elif "86" in outp:
           result = 32
+        else:
+          failMsg = "uname -m = " & outp
+      else:
+        failMsg = "uname -m failed: " & $errC & "\n" & outp
+    else:
+      failMsg = "uname not found"
 
   # Die if unsupported - better fail than guess
-  doAssert result != 0, "getCpuArch() not implemented"
+  doAssert result != 0, "FATAL: Could not detect your CPI architecture\n" & failMsg
 
   # Only once
   cpuArch = result
--- a/src/choosenimpkg/cliparams.nim
+++ b/src/choosenimpkg/cliparams.nim
@@ -149,15 +149,16 @@ proc getCpuArch*(): int =
       failMsg = "uname not found"
 
   # Die if unsupported - better fail than guess
-  doAssert result != 0, "FATAL: Could not detect your CPI architecture\n" & failMsg
+  if result == 0:
+    raise newException(ChooseNimError,
+      "Fatal: Could not detect your CPI architecture\n" & failMsg)
 
   # Only once
   cpuArch = result
 
 proc getMingwPath*(params: CliParams): string =
-  let
-    arch = getCpuArch()
-  return params.getInstallDir() / "mingw" % $arch
+  let arch = getCpuArch()
+  return params.getInstallDir() / "mingw" & $arch
 
 proc getMingwBin*(params: CliParams): string =
   return getMingwPath(params) / "bin"
--- a/src/choosenimpkg/cliparams.nim
+++ b/src/choosenimpkg/cliparams.nim
@@ -1,5 +1,8 @@
 import parseopt, strutils, os
 
+when not defined(windows):
+  import osproc
+
 import nimblepkg/[cli, options, config]
 import nimblepkg/common as nimble_common
 import analytics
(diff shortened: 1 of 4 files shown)


src/choosenimpkg/cliparams.nim
```diff
@@ -99,8 +102,51 @@ proc getCurrentChannelFile*(params: CliParams): string =
...
+      let
+        (outp, errC) = execCmdEx(uname & " -m")
+      if errC == 0:
+        if "64" in outp:
+          result = 64
+        elif "86" in outp:
+          result = 32
+
+  # Die if unsupported - better fail than guess
+  doAssert result != 0, "getCpuArch() not implemented"
```
Araq: But it is implemented, it failed though. How about a real error message like "FATAL: Could not detect your CPU."

genotrance: Updated

src/choosenimpkg/cliparams.nim
```diff
@@ -99,8 +102,62 @@ proc getCurrentChannelFile*(params: CliParams): string =
...
+        if "64" in outp:
+          result = 64
+        elif "86" in outp:
+          result = 32
+        else:
+          failMsg = "uname -m = " & outp
+      else:
+        failMsg = "uname -m failed: " & $errC & "\n" & outp
+    else:
+      failMsg = "uname not found"
```
dom96: What is the purpose of this code? Why not just make this proc windows-only?

src/choosenimpkg/cliparams.nim
```diff
@@ -99,8 +102,62 @@ proc getCurrentChannelFile*(params: CliParams): string =
...
+
+  # Die if unsupported - better fail than guess
+  doAssert result != 0, "FATAL: Could not detect your CPI architecture\n" & failMsg
```
dom96: Do not use asserts to "die". Use normal exceptions, either ``ChooseNimError`` if it's an error that should be shown to the user, or some other exception that should be handled further up the stack.

src/choosenimpkg/cliparams.nim
```diff
@@ -99,8 +102,62 @@ proc getCurrentChannelFile*(params: CliParams): string =
...
+
+  # Only once
+  cpuArch = result
+
 proc getMingwPath*(params: CliParams): string =
-  return params.getInstallDir() / "mingw32"
+  let
+    arch = getCpuArch()
```
dom96: Nit:
```
let arch = getCpuArch()
```

src/choosenimpkg/cliparams.nim
```diff
@@ -99,8 +102,62 @@ proc getCurrentChannelFile*(params: CliParams): string =
...
+  return params.getInstallDir() / "mingw" % $arch
```
dom96: I think this will fail, there are no format strings there (`$1` etc.) or am I missing something?

src/choosenimpkg/cliparams.nim
genotrance: I added it for completeness, if there's a need to detect CPU arch on non-Windows.

src/choosenimpkg/cliparams.nim
genotrance: Agreed.

src/choosenimpkg/cliparams.nim
genotrance: BUG! Fixed.

src/choosenimpkg/cliparams.nim
dom96: meh, just remove it. It's unlikely we will need this and it's bad to keep dead code around.

src/choosenimpkg/cliparams.nim
```diff
@@ -99,8 +102,63 @@ proc getCurrentChannelFile*(params: CliParams): string =
...
+          failMsg = "uname -m = " & outp
+      else:
+        failMsg = "uname -m failed: " & $errC & "\n" & outp
+    else:
+      failMsg = "uname not found"
+
+  # Die if unsupported - better fail than guess
+  if result == 0:
+    raise newException(ChooseNimError,
+      "Fatal: Could not detect your CPU architecture\n" & failMsg)
```
dom96: Nit: you don't need this "Fatal:" prefix in the message.

src/choosenimpkg/cliparams.nim
genotrance: Done

src/choosenimpkg/cliparams.nim
genotrance: Done

dom96:
CI failing? Please also squash your commits (or we can do that when merging the PR).

genotrance:
Okay, CI is finally fixed. I'll squash merge tomorrow if no other feedback.

src/choosenimpkg/cliparams.nim
```diff
@@ -99,8 +99,41 @@ proc getCurrentChannelFile*(params: CliParams): string =
 proc getAnalyticsFile*(params: CliParams): string =
   return params.chooseNimDir / "analytics"
 
+var cpuArch = 0
+
```
dom96: I don't think this deserves memoization, remove this in a follow up PR please.

PR #169, opened 2020-02-01
Add test update self

iranika:
Continuation of # 167
Added test update self in tester.nim .

### DIFF
--- a/tests/tester.nim
+++ b/tests/tester.nim
@@ -176,13 +176,17 @@ when defined(linux):
 test "can update self":
   beginTest()
   cd choosenimpkgDir:
-    copyFile("common.nim", "common.nim.org")
-    writeFile("common.nim", readFile("common.nim").replace(re"chooseNimVersion.*",
+    const commonFile = "common.nim"
+    const commonFileOriginal = "common.nim.org"
+    copyFile(commonFile, commonFileOriginal)
+    writeFile(commonFile, readFile(commonFile).replace(re"chooseNimVersion.*",
                                                   "chooseNimVersion* = \"0.4.0\""))
     cd rootDir:
-      var (output, exitCode) = exec("build", exe="nimble", global=true, liveOutput=true)
+      moveFile(exePath, exePath.addFileExt("org")) # rename Original exe file.
+      var (output, exitCode) = exec("build", exe="nimble", global=false, liveOutput=true)
       check exitCode == QuitSuccess
-    moveFile("common.nim.org", "common.nim")
+    when defined(windows): removeFile(commonFile) # moveFile don't overwritten on windows. So, delete it.
+    moveFile(commonFileOriginal, commonFile)
     
     (output, exitCode) = exec(["update","self","--debug"], liveOutput=true)
     check exitCode == QuitSuccess
@@ -191,4 +195,10 @@ test "can update self":
     (output, exitCode) = exec(["update","self","--debug"], liveOutput=true)
     check exitCode == QuitSuccess
     check inLines(output.processOutput, "Info: Already up to date at version")
+    block cleanup:
+      removeFile(exePath & "0.4.0".addFileExt(ExeExt)) #remove lower version exefile.
+      when defined(windows): removeFile(exePath) # moveFile don't overwritten on windows. So, delete it.
+      moveFile(exePath.addFileExt("org"), exePath) # Return to Original exe file.
+
+  beginTest()  
 
--- a/tests/tester.nim
+++ b/tests/tester.nim
@@ -176,11 +176,6 @@ when defined(linux):
 test "can update self":
   beginTest()
   block :
-    let (output, exitCode) = exec(["update", "self", "--debug"], liveOutput=true)
-    check exitCode == QuitSuccess
-    check inLines(output.processOutput, "Info: Already up to date at version")
-
-  block :
     let (output, exitCode) = exec(["update", "self", "--debug", "--force"], liveOutput=true)
     check exitCode == QuitSuccess
     check inLines(output.processOutput, "Info: Updated choosenim to version")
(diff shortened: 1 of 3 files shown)


tests/tester.nim
```diff
@@ -171,3 +172,31 @@ when defined(linux):
...
+    when defined(windows): removeFile(commonFile) # moveFile don't overwritten on windows. So, delete it.
+    moveFile(commonFileOriginal, commonFile)
+    
+    (output, exitCode) = exec(["update","self","--debug"], liveOutput=true)
+    check exitCode == QuitSuccess
+    check inLines(output.processOutput, "Info: Updated choosenim to version")
+
+    (output, exitCode) = exec(["update","self","--debug"], liveOutput=true)
+    check exitCode == QuitSuccess
+    check inLines(output.processOutput, "Info: Already up to date at version")
```
genotrance: This is checking if the current shipping version of choosenim, that was just updated to, can update. It isn't testing the currently built code. Not sure if there's a point in testing this and if it can break the CI in the future.

tests/tester.nim
```diff
@@ -171,3 +172,31 @@ when defined(linux):
...
+    block cleanup:
+      removeFile(rootDir / "bin" / "choosenim_0.4.0") #remove lower version exefile.
```
genotrance: Where does the `_0.4.0` suffix come from? `nimble build` doesn't append this to the name.

iranika: This is created by updateSelf().
https://github.com/dom96/choosenim/blob/0e7b2c16d4912f508161393d5b1f974eec7d4027/src/choosenim.nim#L92
Old execute files are renamed.

tests/tester.nim
iranika: I should be testing for already updated, before build lower version info.
I don't know risk too, it can break the CI in the future. I want some good ideas.

iranika: Is it better to add a force option to update self?🤔

tests/tester.nim
```diff
@@ -171,3 +172,34 @@ when defined(linux):
       check hasLine(output.processOutput, "switched to nim 1.0.0")
 
       check not dirExists(choosenimDir / "toolchains" / "nim-1.0.0" / "c_code")
+
+test "can update self":
+  beginTest()
+  var (output, exitCode) = exec(["update","self","--debug"], liveOutput=true)
```
dom96: Nit: spaces after comma

iranika:
@dom96 @genotrance 
I added the '--force' option, And rewrote test. I pass Travis test in my branch, But This Pull Reqest failed.
[My branch test result](
https://travis-ci.org/iranika/choosenim/builds/647167264?utm_medium=notification&utm_source=github_status)
[Pull Request test result](https://travis-ci.org/dom96/choosenim/builds/647158838?utm_medium=notification&utm_source=github_status)
I'm not sure now, but I think it is due to `.travis.yml`.  #170
I will examine it later 🤕

genotrance:
@iranika - please hold off on testing this PR, am working on fixing the build script. Also, join us on #nim if possible so that we can chat.

iranika:
@genotrance OK, I hold off on testing this PR, wait # 168 to complete.
I will sleep. Thanks!

tests/tester.nim
```diff
@@ -171,3 +172,21 @@ when defined(linux):
       check hasLine(output.processOutput, "switched to nim 1.0.0")
 
       check not dirExists(choosenimDir / "toolchains" / "nim-1.0.0" / "c_code")
+
+test "can update self":
+  beginTest()
+  block :
+    let (output, exitCode) = exec(["update", "self", "--debug"], liveOutput=true)
+    check exitCode == QuitSuccess
+    check inLines(output.processOutput, "Info: Already up to date at version")
```
dom96: This depends on the external state, which means it will be flaky as soon as we try to run the test suite on an older version of choosenim (which may happen). I think this can be removed.

tests/tester.nim
```diff
@@ -171,3 +172,21 @@ when defined(linux):
...
+  block :
+    let (output, exitCode) = exec(["update", "self", "--debug", "--force"], liveOutput=true)
+    check exitCode == QuitSuccess
+    check inLines(output.processOutput, "Info: Updated choosenim to version")
+    block cleanup:
+      # Now, ExeFile is release version. Back to current build version.
+      # moveFile don't overwritten on windows. So, delete it.
+      when defined(windows): removeFile(exePath)
+      moveFile(rootDir / "bin" / "choosenim_" & chooseNimVersion.addFileExt(ExeExt),
+               exePath)
```
dom96: Why is this needed?

iranika: Because, Test Changes should be undone.
[Line 184](https://github.com/dom96/choosenim/pull/169/files#diff-0e23be51b85eef33ca502065a5cb52e0R184)```let (output, exitCode) = exec(["update", "self", "--debug", "--force"], liveOutput=true)```
This will replace the Exec File  in the bin with the latest `release` version. For subsequent tests, the release version is used, not current build. It's unintended.
And, the artifacts in the bin directory are deployed by TravisCI. In this state, the current build will not be deployed.
So we need back to the state before the test.

tests/tester.nim
iranika: OK, I'll remove it.

tests/tester.nim
iranika: @dom96 I understood why bin dir is used, in this case.
`updateSelf()` doesn't use options `--choosenimDir` and ` --nimbleDir`. It's used getAppFilename().
https://github.com/dom96/choosenim/blob/25f86483552c99c3c70f5271d8a5123f238b9d26/src/choosenim.nim#L91-L102
So, release binary is downloaded to bin dir , in this case.
My mind has changed. Now, I think I should copy `bin/choosenim` to `tests/choosenimDir/` dir, and test using `tests/choosenimDir/choosenim`. In this case, bin dir  has no changed. It's clean. :)
Do you have any thoughts?

dom96: Yeah, I think that sounds better. Although I have a vague memory of how this fits together :)

iranika: @dom96 I done this fix. No trash in the bin dir : )
evidence:
image

PR #170, opened 2020-02-06
Update Nim version, fix travis script - remove commit hash

genotrance:


genotrance:
@dom96 - this is ready to go

PR #173, opened 2020-02-11
Fix #12 - use git to update devel

### LINKED ISSUE #12: Change the way update works
Right now, `choosenim update` just redownloads everything and rebuilds from scratch using csources.

I think it should work this way:

* `choosenim update 0.16.0` (or in fact any tag) should be a noop - there is no need to update a fixed version
* `choosenim update "#head"` should do a git pull and then use koch to rebuild using the existing nim.

In fact, with this behaviour, the version is redundant: the only case in which it makes sense to update is when being on devel, so there could just be `choosenim update` to update #head.

With the current behaviour, choosenim is inconvenient to keep up to date with latest Nim daily - a full clone and rebuild is too slow.

genotrance:
If `devel` or `#head`, and `git` binary is available, fetch updates using `git` and rebuild instead of delete/download/csources/rebuild. Else use legacy method.

Additional [testing](https://travis-ci.org/genotrance/choosenim/builds/648760409) takes much longer, perhaps we don't want it.

cc @Araq

### DIFF
--- a/src/choosenimpkg/download.nim
+++ b/src/choosenimpkg/download.nim
@@ -1,4 +1,4 @@
-import httpclient, strutils, os, terminal, times, json, uri
+import httpclient, strutils, os, osproc, terminal, times, json, uri
 
 when defined(macosx):
   import math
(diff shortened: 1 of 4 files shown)


src/choosenimpkg/download.nim
```diff
@@ -350,6 +350,49 @@ proc getOfficialReleases*(params: CliParams): seq[Version] =
...
+
+      let lastDir = getCurrentDir()
+      setCurrentDir(extractDir)
+      defer:
+        setCurrentDir(lastDir)
+
+      display("Fetching", "latest changes", priority = HighPriority)
+      for cmd in [" fetch --all", " reset --hard origin/devel"]:
+        var (outp, errC) = execCmdEx(git & cmd)
+        if errC != 0:
```
dom96:
```suggestion
        if errC != QuitSuccess:
```

src/choosenimpkg/download.nim
```diff
@@ -350,6 +350,49 @@ proc getOfficialReleases*(params: CliParams): seq[Version] =
...
+        setCurrentDir(lastDir)
+
+      var init = true
+      display("Setting", "up git repository", priority = HighPriority)
+      for cmd in [" init", " remote add origin https://github.com/nim-lang/nim"]:
+        var (outp, errC) = execCmdEx(git & cmd)
+        if errC != 0:
+          display("Warning:", "git" & cmd & " failed: " & outp, Warning, priority = HighPriority)
+          init = false
+          break
```
dom96: Shouldn't this be a fatal error? (i.e. ChoosenimError thrown).

tests/tester.nim
```diff
@@ -171,3 +171,24 @@ when defined(linux):
...
+
+  block:
+    let (output, exitCode) = exec(@["update", "devel"], liveOutput=true)
+    check exitCode == QuitSuccess
+
+    check not inLines(output.processOutput, "extracting")
+    check not inLines(output.processOutput, "setting")
+    check inLines(output.processOutput, "updating")
+    check inLines(output.processOutput, "latest changes")
+    check inLines(output.processOutput, "building")
```
dom96: Nice!

src/choosenimpkg/download.nim
genotrance: If git fails for whatever reason, have already created a .git directory earlier and we default into the legacy way of setting up.

genotrance: Done

src/choosenimpkg/download.nim
genotrance: If git fails for whatever reason, have already created a .git directory earlier and we default into the legacy way of setting up.

PR #174, opened 2020-02-12
Fix #84 - git is requirement

### LINKED ISSUE #84: git is an unlisted requirement
`choosenim` fails when `git` is not in `PATH` on Windows, since `koch` uses `git` to retrieve `nimble` etc.

genotrance:


genotrance:
Doc change, merging.

PR #179, opened 2020-02-17
Install and update nightlies

genotrance:


genotrance:
Devel = #head - untested and has to be compiled from scratch.

Nightly = recent #head - fully tested and precompiled. Nightlies aren't posted if the build fails.

Anyone testing their packages in CI with devel should start using nightlies so that they don't have to waste time rebuilding everything nor be testing against some untested commit. In fact, we won't need to pin Nimble CI testing to a Nim commit anymore, can start using nightlies instead.

Anyone else who likes using bleeding edge will also start using nightlies since choosenim will make it so easy to download and update. They can save their time recompiling everything from scratch and have a fully tested build.

Only true compiler devs will be interested in devel going forward.

genotrance:
Also, `choosenim devel` now uses git to update, recent PR. Nightlies downloads from nim-lang/nightlies.

dom96:
> In fact, we won't need to pin Nimble CI testing to a Nim commit anymore, can start using nightlies instead.

I feel like you're missing the point of pinning commits. We will still need to pin them, as we want reproducible builds. Our CI should be stateless, and using nightlies would introduce state.

----

Anyway, I don't see why anyone would want a `devel` that's broken. I see no reason not to just download the latest nightly for a `devel` request, even if it doesn't equal `devel` at that moment. We should make that clear via installation messages.

genotrance:
Here's my idea of Nim users:

New to Nim
- Uses one release version
- Downloads from nim-lang.org or package manager
- Uses finish
- No need to bootstrap

Regular user
- Uses more than one release of Nim over time
- Uses choosenim x.y.z to install and go back and forth for testing, porting
- No need to bootstrap

Saavy user
- Uses one or more release versions of Nim
- Also tests with bleeding edge to verify stuff works
- Uses choosenim x.y.z + choosenim nightly
- No need to bootstrap

Advanced user
- Uses choosenim devel to checkout code and go around git history
- Wants to git bisect Nim for issues
- Try out PRs
- Can always get back to #head with `choosenim update devel`

Developer
- Can start with `choosenim devel`
- Create a symlink in local dir to `~/.choosenim/toolchains/nim-#devel`
- Use view however they want
- We can add functionality to fork, branch, create PR, etc. using choosenim to make this easy on compiler devs

I want to turn this into a blog post with more details at some point so that people get an idea of what setup they should use to make their life most easy. This needs refinement but gives you an idea of why I think devel != nightlies from an end-user workflow and distinguishing them for users is worth it in choosenim.

cc @Araq @narimiran

dom96:
Developers and advanced users are one and the same. And no, we shouldn't be adding features to help developers of Nim develop Nim... me and Araq have our own checkouts of Nim which we can switch to by running ``choosenim ~/git/nim`` (or whatever path our checkout is in). This is why I don't think having a devel/nightly split makes sense.

That and also the fact that in 99% of instances the nightlies will only be a little behind `devel`. To the end user this small lag does not matter.

iranika:
I think it’s better to add a description to the Example.
https://github.com/dom96/choosenim/blob/b231093ff7f167c0adbc884cbecba75f27d23f0b/src/choosenimpkg/cliparams.nim#L38-L39

genotrance:
To me, #head is different from nightlies and both have different target audiences. I wish others chimed in too but I know I'd use nightlies in CI for sure, whereas use `choosenim devel` when wanting to test against #head locally. Running `choosenim update devel` is much faster than doing the git/koch dance.

By retaining this separation, we can further add functionality to [test PRs](https://github.com/dom96/choosenim/issues/58) and branches and make life easier. All that is out if nightlies == devel.

Anyone working on Nim features will be working on a fork in a local directory. Maybe that never ends up beyond `choosenim path/to/nimfork` but choosenim could still improve life there. I don't see why we cannot explore making that workflow better. There is no value in merging two separate user workflows especially when the code already exists and opens doors for improvements.

Anyway, if you don't think this separation is warranted, let's just close this PR.

dom96:
> I wish others chimed in too but I know I'd use nightlies in CI for sure, whereas use choosenim devel when wanting to test against #head locally. Running choosenim update devel is much faster than doing the git/koch dance.

I don't understand the point, why test specifically against `#head` locally? Why not just test against nightlies? Again, in 99% of cases #head will be the same as your `nightly` concept. I just don't see the point in separating them.

The only reason I can see is if your assumption is that nightly builds will be down for long periods of time. But is that the case in practice?

genotrance:
I just used `choosenim update devel` - https://github.com/nim-lang/Nim/pull/13472 is not passing CI due to a broken test in #head and I wanted to check if it was fixed in the latest commit. Nightlies is 1 commit behind though it can routinely be further behind depending on how active the day is.

Araq:
>  But is that the case in practice?

Yes but we're investigating why and hope to fix this problem for good.

dom96:
Okay, so how about this:

* Default behaviour for `choosenim update devel`: grabs latest nightly
  * but detects if this matches `#devel`'s latest commit, if not it prints a warning (with a suggestion to use below flag)
* `choosenim update devel --ignoreNightly` (or something) to force choosenim to build latest

So essentially my main concern here is that `nightly` shouldn't be its own channel. Hope that makes sense.

Araq:
> Anyway, if you don't think this separation is warranted, let's just close this PR.

If in doubt, leave it out. Remember that we need to support all these different modes and combinations effectively until forever.

genotrance:
I've updated this PR as follows:-

`choosenim devel` = Install latest nightly

`choosenim update devel` =  Update to latest nightly

`choosenim devel --latest` = Download and bootstrap latest commit

`choosenim update devel --latest` = Update via git and rebuild

Effectively, legacy behavior of `choosenim [update] devel` is now `choosenim [update] devel --latest` and default devel behavior is to simply use nightlies.

PTAL

genotrance:
Test is failing because we haven't had a Windows nightly build since 1/23 and it does not show up on page 1 of the releases list. I have added a fallback to the legacy method if this happens.

src/choosenimpkg/cliparams.nim
```diff
@@ -32,13 +32,16 @@ Example:
   choosenim stable
     Installs (if necessary) Nim from the stable channel (latest stable release)
     and then selects it.
-  choosenim #head
-    Installs (if necessary) and selects the latest current commit of Nim.
-    Warning: Your shell may need quotes around `#head`: choosenim "#head".
+  choosenim devel [--latest]
+    Installs (if necessary) and selects the most recent nightly build of Nim.
+    The '--latest' flag selects the current commit of Nim.
```
dom96:
```suggestion
    The '--latest' flag selects and builds the latest commit in the devel branch
```

src/choosenimpkg/cliparams.nim
```diff
@@ -32,13 +32,16 @@ Example:
...
   choosenim ~/projects/nim
     Selects the specified Nim installation.
   choosenim update stable
     Updates the version installed on the stable release channel.
+  choosenim update devel [--latest]
+    Updates to the most recent nightly build of Nim.
+    The '--latest' flag updates to the current commit of Nim.
```
dom96: Change in the same way as above.

src/choosenimpkg/download.nim
```diff
@@ -11,6 +11,7 @@ import cliparams, common, telemetry, utils
 
 const
   githubTagReleasesUrl = "https://api.github.com/repos/nim-lang/Nim/tags"
+  githubNightliesReleasesUrl = "https://api.github.com/repos/nim-lang/nightlies/releases"
```
dom96: Is there any way to get this without using the GitHub API? The API is rate limited to 60 request per hour (and I bet that can change in the future) so users may get rate limited (especially if they are developing GitHub API apps or other apps use this limit).

src/choosenimpkg/download.nim
```diff
@@ -206,21 +207,51 @@ proc needsDownload(params: CliParams, downloadUrl: string,
...
+            let aname = asset["name"].getStr()
+            if os in aname:
+              when not defined(macosx):
+                if "x" & $arch in aname:
+                  url = asset["browser_download_url"].getStr()
+              else:
+                url = asset["browser_download_url"].getStr()
+            if url.len != 0:
+              break
+        if url.len != 0:
```
dom96: Refactor this parsing logic into a separate function

src/choosenimpkg/download.nim
```diff
@@ -206,21 +207,51 @@ proc needsDownload(params: CliParams, downloadUrl: string,
...
+      let os =
+        when defined(windows): "windows"
+        elif defined(linux): "linux"
+        elif defined(macosx): "osx"
+      reference = "devel"
+      for jn in parsedContents.getElems():
+        if jn["name"].getStr().contains("devel"):
+          for asset in jn["assets"].getElems():
+            let aname = asset["name"].getStr()
+            if os in aname:
```
dom96: why do we need to do this? Can't we do full string matching instead of substring?

tests/tester.nim
```diff
@@ -172,26 +172,49 @@ when defined(linux):
...
-    check exitCode == QuitSuccess
+    # Travis runs into Github API limit
+    if not inLines(output.processOutput, "403"):
+      check exitCode == QuitSuccess
 
-    check not inLines(output.processOutput, "extracting")
-    check not inLines(output.processOutput, "setting")
-    check inLines(output.processOutput, "updating")
-    check inLines(output.processOutput, "latest changes")
-    check inLines(output.processOutput, "building")
```
dom96: Please keep these tests, just add ``--latest`` to them to ensure you didn't break existing functionality.

src/choosenimpkg/cliparams.nim
genotrance: Done

src/choosenimpkg/cliparams.nim
genotrance: Done

src/choosenimpkg/download.nim
genotrance: Agreed but I cannot think of any other way besides scraping the website.

src/choosenimpkg/download.nim
genotrance: `asset["name"]` is the filename, like `nim-1.1.1-osx.tar.xz` or `nim-1.1.1-linux_x64.tar.xz` or `nim-1.0.7-windows_x64.zip`. Have to look for OS within filename.

src/choosenimpkg/download.nim
genotrance: Done

tests/tester.nim
genotrance: Testing will be much longer but sure.

PR #187, opened 2020-03-24
Support wget

ansimita:
Tested successfully on Debian Buster without `curl` installed.

To check whether a command exists, I used the code provided in [this StackOverflow answer](https://stackoverflow.com/a/677212).

For your convenience: the [wget(1) man page](https://linux.die.net/man/1/wget).

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -39,6 +39,9 @@ prompts and enjoy your new Nim and choosenim installation.
 ```
 curl https://nim-lang.org/choosenim/init.sh -sSf | sh
 ```
+```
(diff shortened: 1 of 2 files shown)


ansimita:
Yes. Aside from the minimal installation of Debian, both the [standard and extended editions](https://alpinelinux.org/downloads/) of Alpine Linux rely on [busybox wget](https://busybox.net/downloads/BusyBox.html#wget) and does come with curl preinstalled.

genotrance:
Looks good to me. @dom96?

PR #201, opened 2020-05-24
Fix #199 - missing dlls

### LINKED ISSUE #199: Nimarchive cannot overwrite existing extraction?
Getting this error which makes choosenim unusable:

```
PS C:\Users\morfe\projects\stardust> choosenim 1.2.0
      Info: C:\Users\morfe\.choosenim\downloads\dlls.zip already downloaded
 Extracting dlls.zip
     Error: Unable to extract. Error was 'Cannot create a file when that file already exists.
        ... '.
```

The extraction is into nimble bin dir (the same as returned by ``choosenim --getNimbleBin``). So it fails because it's trying to extract DLLs that are already there...

Steps to repro:

* Rename libeay32.dll to libeay32.dll.old (or any other dll installed by choosenim)
* Run `choosenim 1.2.0`

You will see that error.

genotrance:


genotrance:
@dom96 any change of heart?

https://github.com/dom96/choosenim/issues/199#issuecomment-632762684

dom96:
Let's do this, show a warning in two situations:

* if certain DLLs are missing and have been copied into ~/.nimble/bin (one message for each DLL that was copied)
* if a DLL already exists, but a new version has been installed (one for each DLL that was changed).

Does that sound reasonable?

genotrance:
Okay, all passing - PTAL.

src/choosenim.nim
```diff
@@ -56,8 +58,25 @@ proc chooseVersion(version: string, params: CliParams) =
...
+          try:
+            if not fileExists(binDir / path) or
+              getLastModificationTime(binDir / path) < getLastModificationTime(tempDir / path):
+              moveFile(tempDir / path, binDir / path)
+              display("Info:", "Copied '$1' to '$2'" % [path, binDir], priority = HighPriority)
+          except:
+            discard
+      while dirExists(tempDir):
+        removeDir(tempDir)
+        sleep(500)
```
dom96, line 79: Wat?

PR #202, opened 2020-05-29
Create Windows setup file

ArMour85:
Hello

My goal is to create an installation file for Windows with mingw. It will allow developers to install easily without using a terminal and ready to code.

If and when this PR will be accepted I will propose this installer in the Windows package manager aka winget.

Unfortunately I don't know Travis and I can't test the CI, I hope that it isn't  too complicate.


ArMour85:
I'll create a version without mingw.

ArMour85:
Hello @dom96,

Can you say me if you can test my installer exe and if this PR is ok for you ?

Thanks.

scripts/inno-choosenim-setup.iss
```diff
@@ -0,0 +1,115 @@
+; Script generated by the Inno Setup Script Wizard.
+; SEE THE DOCUMENTATION FOR DETAILS ON CREATING INNO SETUP SCRIPT FILES!
+
+#define MyAppName "choosenim"
+#define MyAppVersion "0.6.0"
```
dom96, line 5: Having to update this manually sucks, any way to not have to do that?

ArMour85: I began my work in **niminst**  but there some issues. I think it isn't used or discarded.

I think that **choosenim** is the best approach to install and manage language. Some languages are the same approach like **RUST** with **rustup** and **Julia** with **Jill**.

ArMour85: > Having to update this manually sucks, any way to not have to do that?

Yes of course, I work in it tonight.

ArMour85:
It's done, but there is a problem with **yes | choosenim stable** command during CI.
```
$ yes | choosenim stable
      Info: C:\Users\travis\.choosenim\downloads\dlls.zip already downloaded
 Extracting dlls.zip
       Tip: 4 messages have been suppressed, use --verbose to show them.
     Error: Unable to extract. Error was 'Cannot create a file when that file already exists.
        ... '.
The command "yes | choosenim stable" exited with 1.
```

ArMour85:
I don't understand this problem for Windows. In your code you use "-o" option to overwrite extracted files.

ArMour85:
@dom96 can you run the CI on the master branch to check if it's my modifications that crash the CI ?
Please.

ArMourMag:
The last fix in master has resolved my CI issue.

ArMour85:
The last fix in master has resolved my CI issue.

ArMour85:
It's done ! I removed the MinGW version because choosenim download it if missing.

@dom96, can you merge this branch please ?

Also, where is zip archive is created ? I needed to copy dlls and I didn't find the zip process as example.

Thanks.

ArMour85:
@genotrance and @dom96, can you try the setup in your computer before validate please ?
I test only on my computer and It will be interesting to try in yours.

genotrance:
@ArMour85 - as great as this is, is there any value given choosenim can now standalone on Windows (no dlls) and that it can also add to PATH with --firstInstall? What else does this installer do for the user?

ArMour85:
@genotrance, when you want to install an application in Windows and you have the choice between a zip file and an exe file, which one do you use?
Certainly the exe file because it is simpler, you do "next->next->close". It's the same for Linux or MacOS, an installer will always be preferred.

It is a mistake to believe that a developer is necessarily a computer scientist. I know some very good mathematicians who can write super algorithms in Python but who don't know how to use a terminal and don't know the environment variables. They use the installation exe and run the Spider IDE.
I once asked a colleague to test the Julia language for his work and he answered that he will do it when an installer is proposed but not before :/

On Nim's forum we often talk about how to increase the popularity of the language, forgetting that it must first be installable everywhere and easily. I consider choosenim to be the official installer of the Nim language in the same way as Rustup (https://www.rust-lang.org/learn/get-started).

To come back to your question, the installer will do the same thing as the "runme.bat" script but above all it will create an uninstaller that will delete the folders created by choosenim. Which is for me a minimum.

If this PR is accepted, I will publish the installer on winget (https://github.com/microsoft/winget-cli).

dom96:
I feel like this overcomplicates something that we've already had. We used to provide a self-extracting archive which works in a very similar way to the rustup-init.exe, I think we could start doing this again. 

Thinking about this some more, maintaining this would be a pain, we've been there and done that with Nim itself. The only thing missing from the solution I outline above is a way to uninstall choosenim, but I bet there is an easier solution here too, perhaps we can simply have choosenim create a deinstaller exe on Windows machines and link to it from whatever registry necessary for it to show up in Windows' Add/Remove Programs list?

genotrance:
I think the whole point of this PR is to make it accessible to those who are used to having a GUI for install purposes. No harm posting this for every build.

genotrance:
I moved build.sh into the scripts directory, might need to rebase for that. Also, will need to use $VERSION_TAG instead of $VERSION.

ArMour85:
Last changes done.

ArMour85:
Do you think you will soon have time to merge it ?

dom96:
Needs rebase. I will merge this, but a word of warning, if it's broken or breaks the build I will not wait long before removing it.

ArMour85:
Nobody can guarantee that will work forever. If there are modifications that break the build generation then the cause will be in the `build.sh` script but not in Inno Setup script.

If use Inno Setup  gives you a rash and you are afraid of your build process. We can stop the merge request and I will create a external project to generate installer for Windows, Linux (deb and rpm) and MacOs.

It's you the maintainer, you decide.

ArMour85:
I decided to create an independent project to create the Windows setup [here](https://gitlab.com/ArMour85/choosenim-setup). The project is hosted by GitLab.

Like that, no problem of broken CI and you have nothing to do.

This project's CI, every week, download the latest release and create the Windows setup if it doesn't exist. I also made a Chocolatey package that is in validation process.

A little link in the readme will be very appreciate.

And Happy New Year 🍾

dom96:
Sounds good, thank you! Added to the readme. Happy new year :)

PR #213, opened 2020-08-18
Support linux_arm64 and linux_powerpc64

elijahr:
This will be especially useful for folks who want to test their Nim projects across all platforms on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,17 +24,21 @@ install:
   # https://gist.github.com/genotrance/fb53504a4fba88bc5201d3783df5c522
   - curl https://gist.github.com/genotrance/fb53504a4fba88bc5201d3783df5c522/raw/travis.sh -LsSf -o travis.sh
   - |
-    build_nim () {
-      pushd ~
-      git clone -b v${BRANCH} --single-branch https://github.com/nim-lang/Nim.git
-      pushd Nim
-      sh build_all.sh
-      popd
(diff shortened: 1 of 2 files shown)


dom96:
Hrm, not sure why but the CI didn't run for this PR.

elijahr:
@dom96 The error ([arm64](https://travis-ci.org/github/dom96/choosenim/jobs/719352905), [ppc64](https://travis-ci.org/github/dom96/choosenim/jobs/718869211)) on Travis is:

```
Failed to download libiconv v1.16 from Conan - check https://conan.io/center [AssertionError]
```

[Conan](https://conan.io/center/libiconv/1.16/?user=_&channel=_&revision=&os=Linux&tab=config) does indeed show that there are no arm64 or ppc64 libiconv binaries. I'm not yet familiar with nimterop, but I suspect that it's trying to fetch the libiconv binaries from Conan as a fallback because it fails to find them on the host. Probably just requires some additional package installation with `apt-get`, but might be an upstream bug somewhere. I'll have some time to investigate further in the next week or so.

genotrance:
Yes, conan doesn't have arm64 or ppc64 libs. Neither does binarybuilder.org so best option is to apt-get.

elijahr:
Closing this in favor of some more in-depth work to support these architectures, will be submitting a PR shortly.

PR #218, opened 2020-09-05
[FIX(algo)] Generate correct error message

haxscramper:
Generate correct error message on missing C compiler based on target
OS - `clang` for OSX and `gcc` for linux.

### DIFF
--- a/src/choosenim.nim
+++ b/src/choosenim.nim
@@ -50,9 +50,17 @@ proc chooseVersion(version: string, params: CliParams) =
       let path = downloadMingw(params)
       extract(path, getMingwPath(params))
     else:
-      raise newException(ChooseNimError,
-                         "No C compiler found. Nim compiler requires a C compiler.\n" &
-                         "Install clang or gcc using your favourite package manager.")
+      let binName =
+        when defined(OSX):
+          "clang"
+        else:
+          "gcc"
+
+      raise newException(
+        ChooseNimError,
+        "No C compiler found. Nim compiler requires a C compiler.\n" &
+        "Install " & binName & " using your favourite package manager."
+      )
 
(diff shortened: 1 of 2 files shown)


genotrance:
Why not instead update isDefaultCCInPath to check clang on Linux and Windows instead?

haxscramper:
I was told it is default Nim behavior to assume that GCC is present on Linux and clang on osx. 

Some other code might depend on this assumption and I'm not exactly sure if this change in behavior won't break anything. 

If using clang won't break anything I will of course change check to include clang too.

Additional question: maybe checking for tcc us possible too, in cases where Nim compiler is available as a binary download for platform? AFAIK it cannot be built using tcc due to some errors (not confirmed, will check later).

src/choosenim.nim
```diff
@@ -50,9 +50,17 @@ proc chooseVersion(version: string, params: CliParams) =
       let path = downloadMingw(params)
       extract(path, getMingwPath(params))
     else:
-      raise newException(ChooseNimError,
-                         "No C compiler found. Nim compiler requires a C compiler.\n" &
-                         "Install clang or gcc using your favourite package manager.")
+      let binName =
+        when defined(OSX):
```
dom96, line 54:
```suggestion
        when defined(macosx):
```

dom96 (approved):
Seems fine to me, although I'm not sure `OSX` is correct (and we use `macosx` everywhere else in choosenim...)

haxscramper:
`isDefaultCCInPath` also used `when defined(OSX)` - fixed it too.

haxscramper:
Adding check for `clang` or `gcc` (and `tcc`) is possible or it might break some other logic? 

I'm talking about cases where pre-built download of nim binaries is used & user wants to have clang/tcc installed.

PR #226, opened 2020-10-07
Nightly update improvements

dom96:
We now get the tag name in the "Downloading" message:

image

Plus the behaviour of `update` changed (see changelog).

**Future improvements:** we should have ``download.downloadImpl`` return ``(string, Version)``, the latter being the version that has been downloaded. Sadly it looks like we'll have to parse out the commit from the tag name to get this for the nightlies. This is a minor thing, but it will allow us to display "Updated to \<commit hash\>" rather than "Updated to #devel".

### DIFF
--- a/changelog.markdown
+++ b/changelog.markdown
@@ -1,12 +1,21 @@
 # Choosenim changelog
 
-## 0.6.2 - xx/xx/2020
+## 0.7.0 - 16/10/2020
+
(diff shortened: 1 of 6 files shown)


genotrance:
Might need to rebase to latest choosenim - I'm not sure which package is asking for nim 1.2.6 if you haven't rebased.

PR #228, opened 2020-10-10
Fix #123 - add remove command

### LINKED ISSUE #123: Add a command to remove versions
Currently, there doesn't seem to be a command like `choosenim remove <version>` to remove an installed version. Removing versions can still be done manually by removing directories in `$HOME/.choosenim/toolchains/`, but having a built-in command would be more user-friendly.

This command should probably only work on versions downloaded by choosenim, not on self-provided versions.

EvilKhaosKat:
Added new `remove` command and a test for it.
Unfortunately for some reasons tests in general didn't work at my current installation, so I wasn't able to check whether the test is actually working.
Though tested the command on myself and removed a couple of old nim versions :)

### DIFF
--- a/src/choosenim.nim
+++ b/src/choosenim.nim
@@ -274,6 +274,28 @@ proc versions(params: CliParams) =
         display("", $version & isLatestTag(params, version), priority = HighPriority)
     echo ""
 
+proc remove(params: CliParams) =
+  if params.commands.len != 2:
+    raise newException(ChooseNimError,
+                       "Expected 1 parameter to 'remove' command")
+
+  let version = params.commands[1].newVersion
+
+  let isInstalled = isVersionInstalled(params, version)
+  if not isInstalled:
+    raise newException(ChooseNimError,
+                       "Version $1 is not installed." % $version)
+
+  display("Removing", $version, priority = HighPriority)
+
+  let extractDir = params.getInstallationDir(version)
+  removeDir(extractDir)
+
+  display("Info:", "Removed version " & $version,
+          Success, HighPriority)
+
+  # TODO: switch to latest available version if current was removed
+
 proc performAction(params: CliParams) =
   # Report telemetry.
   report(initEvent(ActionEvent), params)
@@ -285,6 +307,8 @@ proc performAction(params: CliParams) =
     show(params)
   of "versions":
     versions(params)
+  of "remove":
+    remove(params)
   else:
     choose(params)
 
--- a/src/choosenim.nim
+++ b/src/choosenim.nim
@@ -286,6 +286,7 @@ proc remove(params: CliParams) =
     raise newException(ChooseNimError,
                        "Version $1 is not installed." % $version)
 
+  let isCurrentVersionRemove = version == getCurrentVersion(params)
   display("Removing", $version, priority = HighPriority)
 
   let extractDir = params.getInstallationDir(version)
@@ -293,6 +294,16 @@ proc remove(params: CliParams) =
 
   display("Info:", "Removed version " & $version,
           Success, HighPriority)
+
+  if isCurrentVersionRemove:
+    let installedVersions = getInstalledVersions(params)
+    if installedVersions.len > 0:
+      let versionToUse = installedVersions[0]
+      display("Warning:", "Removed version was selected, replace with " & $versionToUse, Warning, priority = HighPriority)
+      switchTo(versionToUse, params)
+    else:
+      display("Warning:", "Removed version was selected", Warning, priority = HighPriority)
+
 
   # TODO: switch to latest available version if current was removed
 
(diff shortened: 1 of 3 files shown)


tests/tester.nim
```diff
@@ -248,3 +248,24 @@ test "can update self":
...
+#       check inLines(output.processOutput, "already built")
+#     else:
+#       check inLines(output.processOutput, "building tools")
+#     check hasLine(output.processOutput, "switched to nim 0.16.0")
+
+#   block:
+#     let (output, exitCode) = exec("remove 0.16.0")
+#     check exitCode == QuitSuccess
+
+#     check hasLine(output.processOutput, "Info: Removing 0.16.0")
```
dom96, line 271: Where does this fail? Can you enable it so that the CI can run it and give some hints? :)

src/choosenim.nim
```diff
@@ -274,6 +274,28 @@ proc versions(params: CliParams) =
...
+
+  display("Removing", $version, priority = HighPriority)
+
+  let extractDir = params.getInstallationDir(version)
+  removeDir(extractDir)
+
+  display("Info:", "Removed version " & $version,
+          Success, HighPriority)
+
+  # TODO: switch to latest available version if current was removed
```
dom96, line 297: Would be brilliant if you could implement this too. Otherwise the user may get odd errors if they remove a version they have currently active.

dom96 (approved):
Nice! Thanks for creating this!

src/choosenim.nim
EvilKhaosKat: Sure, I'll give it a try :)

tests/tester.nim
EvilKhaosKat: >Where does this fail?
For me it fails on several tests, for example "can choose v0.16.0", with `koch.nim(24, 3) Error: cannot open 'os'`. Probably some mac~ish issue

> Can you enable it so that the CI can run it and give some hints? :)
I'll try this out :)

src/choosenim.nim
EvilKhaosKat: Added

By the way - could it be useful to prohibit removal of the last installed version? (on the other hand - iirc `choosenim` doesn't install nim by default during it's own installation, so zero installed versions is a valid case)

tests/tester.nim
genotrance: Please just remove a version installed in a previous test - bootstrapping 0.16.0 again will add a lot of time to the CI.

src/choosenim.nim
genotrance: Why not just prevent removal of current selected version? Will be simpler.

EvilKhaosKat: Sounds good, will update the approach then

EvilKhaosKat: Added

tests/tester.nim
EvilKhaosKat: Skipped the cleanup stage for the test to have version to remove from previous test.
But considering further mentioned feature 'prohibit current version removal' - I rewrote tests to 'can't remove current version' and 'can't remove non-existent version'

src/choosenim.nim
```diff
@@ -274,6 +274,31 @@ proc versions(params: CliParams) =
...
+
+  let extractDir = params.getInstallationDir(version)
+  removeDir(extractDir)
+
+  display("Info:", "Removed version " & $version,
+          Success, HighPriority)
+
+
+  # TODO: switch to latest available version if current was removed
+
```
dom96, line 301: Nit: remove this comment :)

src/choosenimpkg/cliparams.nim
```diff
@@ -44,6 +44,8 @@ Example:
     The '--latest' flag updates and builds the latest commit in the devel branch
   choosenim versions [--installed]
     Lists the available versions of Nim that choosenim has access to.
+  choosenim remove 0.16.0
+    Removes (if installed) version 0.16.0 of Nim.
```
dom96, line 48: Nit: 0.16.0 is old, let's show examples with more common versions like 1.0.6 :)
```suggestion
  choosenim remove 1.0.6
    Removes (if installed) version 1.0.6 of Nim.
```

tests/tester.nim
```diff
@@ -159,6 +159,23 @@ test "can choose v0.16.0":
     check exitCode == QuitSuccess
     check inLines(output.processOutput, "v0.8.2")
 
+test "cannot remove current v0.16.0":
+  # skip prev test cleanup to have version to remove
+  #beginTest()
```
dom96, line 164: Having tests depend on other tests is in general bad, but I'll let it slide this time. Maybe just add a TODO here to fix this?

dom96 (approved):
Looks great. Just a few tiny things and we can merge.

src/choosenim.nim
EvilKhaosKat: Oops :)

tests/tester.nim
EvilKhaosKat: Yeah, good idea. Added reason _why_ we rely on previous test here + TODO for fixing

PR #235, opened 2020-10-19
Fix nimble issue 864

genotrance:


choosenim.nimble
```diff
@@ -1,6 +1,6 @@
 # Package
 
-version       = "0.7.2"
+version       = "0.7.3"
```
dom96, line 3: 0.7.4. I know we weren't consistent with this, but I went straight to 0.7.2 so...

src/choosenimpkg/cliparams.nim
```diff
@@ -176,7 +176,7 @@ proc newCliParams*(proxyExeMode: bool): CliParams =
     result.nimbleOptions = initOptions()
     if not proxyExeMode:
       result.nimbleOptions.config = parseConfig()
-    setNimbleDir(result.nimbleOptions)
+      setNimbleDir(result.nimbleOptions)
```
dom96, line 179: damn, I misread this condition when I wrote this.

dom96 (approved):

choosenim.nimble
genotrance: You were, I never noticed. I'll bump to 0.7.4 if we want to ship this right away. Else we can leave it as a devel version until we are ready to tag.

PR #273, opened 2021-10-04
add extra feature "choosenim show path"

### LINKED ISSUE #260: "choosenim show path" command
Some projects (like [napibindings](https://github.com/AjBreidenbach/napibindings)) need a command that prints only the path of the current Nim version.

```
$ choosenim show path
/Users/foo/.choosenim/toolchains/nim-1.4.8
```

ilmanzo:
fix #260

### DIFF
--- a/src/choosenim.nim
+++ b/src/choosenim.nim
@@ -207,7 +207,7 @@ proc show(params: CliParams) =
   let channel = getCurrentChannel(params)
   let path = getSelectedPath(params)
   let (_, version) = getNameVersion(path)
-  if params.commands.len==2:
+  if params.commands.len == 2:
     let whatToShow=params.commands[1]
     if whatToShow.toLowerAscii()=="path":
       echo path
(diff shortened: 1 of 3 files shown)


src/choosenim.nim
```diff
@@ -207,6 +207,11 @@ proc show(params: CliParams) =
   let channel = getCurrentChannel(params)
   let path = getSelectedPath(params)
   let (_, version) = getNameVersion(path)
+  if params.commands.len==2:
```
dom96, line 210:
```suggestion
  if params.commands.len == 2:
```

src/choosenim.nim
```diff
@@ -207,6 +207,11 @@ proc show(params: CliParams) =
...
+    let whatToShow=params.commands[1]
```
dom96, line 211:
```suggestion
    let whatToShow = params.commands[1]
```

src/choosenim.nim
```diff
@@ -207,6 +207,11 @@ proc show(params: CliParams) =
...
+    if whatToShow.toLowerAscii()=="path":
```
dom96, line 212:
```suggestion
    if whatToShow.toLowerAscii() == "path":
```

dom96 (approved):
Nice, thanks! If you have time for another PR: a test for this would be awesome.

ilmanzo:
> Nice, thanks! If you have time for another PR: a test for this would be awesome.

sure, please take a look at #276

PR #277, opened 2021-10-07
Error on 404

ire4ever1190:
Checks the `Response` objects http code and returns error if 404 instead of writing an empty tar file and trying to extract that.

Right now it only checks for 404, should it return an error unless the http code is 2xx?


ire4ever1190:
> but really this should be fixed in puppy too

This is by design in puppy as shown in the readme. I have left a question in the puppy repo asking if the exceptions could be thrown when using the basic proc (https://github.com/treeform/puppy/issues/31)

PR #279, opened 2021-10-16
fix: run choosenim update stable before downloading nim

### LINKED ISSUE #264: Possible issue detecting older choosenim install
Running choosenim using the `curl` command installed the older stable version (1.2.0 instead of 1.4.8).

Perhaps since there was an older choosenim install already, the script was somehow unable to detect the older install.

This was resolved for now by running `choosenim update stable`, however the author requested that an issue be filed to look into on the older-install check.

ganderzz:
Resolves  #264 

Run `choosenim update stable` before downloading and installing nim.
This prevents a local `stable` pointing to an old version of nim.`

### DIFF
--- a/scripts/choosenim-unix-init.sh
+++ b/scripts/choosenim-unix-init.sh
@@ -78,6 +78,7 @@ install() {
     # Install Nim from desired channel.
     "$temp_prefix/$filename" $CHOOSE_VERSION --firstInstall ${debug} < /dev/tty
   else
+    "$temp_prefix/$filename" update $CHOOSE_VERSION -y
     "$temp_prefix/$filename" $CHOOSE_VERSION --firstInstall -y ${debug}
   fi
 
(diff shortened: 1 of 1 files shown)


dom96 (approved):
In general the right idea, but see my comments. Thank you for working on this!

dom96:
So I think you need it in both branches. I'd merge this but this needs testing, can you test it with choosenim not installed + choosenim installed and include the output you get? :)

ganderzz:
@dom96  Updated. :)

**With choosenim installed**
Screen Shot 2021-11-09 at 6 51 52 PM

**Without**

Screen Shot 2021-11-09 at 6 54 20 PM

dom96 (changes requested):
Sorry, but thinking about this some more I think this logic should be implemented inside choosenim. i.e. if `choosenim stable` is ran with the `--firstInstall` flag then run `update` implicitly. That way it's easier to keep this logic sound.

PR #286, opened 2021-11-14
fix init.sh script to use x86_64 binaries for ARM macs

100phlecs:
Resolves #283 :)


dom96:
bah, that's it huh. Thanks!

dom96:
Deployed online too so if you run the curl line it should work now :)

Issue #149, opened 2019-12-14
Add ARM64 support


mratsim:
Similar to #43

Travis now supports ARM64 in CI (https://travis-ci.com/mratsim/Synthesis/jobs/267288652#L55), I usually use choosenim to manage my CI runs but unfortunately that would prevent me from using it for CI on ARM.

sk-Prime:
Please support arm architecture. raspberry pi nano, armv6

wantMoreSleep:
And also Apple Silicone chip (M1) (macosx_arm)

p0w1nd:
I've just bought a Macbook Pro M1 and disappointed to see this error when installing choosenim:

```
$ curl https://nim-lang.org/choosenim/init.sh -sSf | sh

choosenim-init: Error: Sorry, your platform (macosx_arm) is not supported by choosenim.
choosenim-init: Error: You will need to install Nim using an alternative method.
choosenim-init: Error: See the following link for more info: https://nim-lang.org/install.html
```

Hope to see Apple M1 support soon.

dom96:
@p0w1nd see https://github.com/dom96/choosenim/pull/283#issuecomment-949053963. Support should be fairly easy by just modifying the bash script. If you can help I would appreciate it :)

dleslie:
FYI, this is breaking on all ARM targets; including Raspbian and Armbian.

adokitkat:
Bumping this. Nim nightlies are being pre-build for ARM64 (and also ARM7L).

bung87:
track issue: https://github.com/github/roadmap/issues/528

DrorDvash:
@dom96 running a linux vm on Apple Silicone chip (m1/m2) is actually aarch64 or linux_arm64.
```
# uname -m      
aarch64
```
Running choosenim will end up with this error:
`choosenim-init: Error: Sorry, your platform (linux_arm64) is not supported by choosenim.`

anyway, both are the same and seem to be unsupported. is there any workaround/fix for this?

vandot:
M1 runners available https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

PR #292, opened 2022-01-26
build arm64 even on rosetta ... improve #149

heinthanth:
On M1 macs, since we don't have arm binaries for choosenim, we can use x86 choosenim to download and invoke compiler. By default choosenim under rosetta build x86 nim. But passing arch -arm64 will build arm64 nim binary. This add improvement to #149.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,13 @@ os:
   - osx
   - windows
 
+dist: bionic
+osx_image: xcode12.3
+
 language: c
 
 env:
-  - BRANCH=1.2.8
+  - BRANCH=1.4.4
 
 cache:
   directories:
--- a/src/choosenimpkg/utils.nim
+++ b/src/choosenimpkg/utils.nim
@@ -30,25 +30,14 @@ proc parseVersion*(versionStr: string): Version =
 
   result = newVersion(versionStr)
 
-proc isRosetta(): bool =
-  let res = execCmdEx("sysctl -in sysctl.proc_translated")
-  if res.exitCode == 0:
-    return res.output.strip() == "1"
-  return false
-
 proc doCmdRaw*(cmd: string) =
-  var command = cmd
   # To keep output in sequence
   stdout.flushFile()
   stderr.flushFile()
 
-  if defined(macosx) and isRosetta():
-    command = "arch -arm64 " & cmd
-
-  displayDebug("Executing", command)
+  displayDebug("Executing", cmd)
   displayDebug("Work Dir", getCurrentDir())
-
-  let (output, exitCode) = execCmdEx(command)
+  let (output, exitCode) = execCmdEx(cmd)
   displayDebug("Finished", "with exit code " & $exitCode)
   displayDebug("Output", output)
 
--- a/src/choosenimpkg/utils.nim
+++ b/src/choosenimpkg/utils.nim
@@ -30,14 +30,25 @@ proc parseVersion*(versionStr: string): Version =
 
   result = newVersion(versionStr)
 
+proc isRosetta(): bool =
+  let res = execCmdEx("sysctl -in sysctl.proc_translated")
+  if res.exitCode == 0:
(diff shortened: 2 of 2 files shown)


heinthanth:
@dom96 can you check it 😄

heinthanth:
Hum ... quite interesting ... need to figure out ... ( So, this PR still not ready to merge )

Screen Shot 2022-01-26 at 8 34 59 PM

heinthanth:
`arm64 nim c koch` generate `x86_64 koch`

Screen Shot 2022-01-26 at 8 44 20 PM

heinthanth:
Working on it :P

dom96 (commented):
This is a cool approach, thank you!

I'm not sure we can just do it for every command though. Can we limit it to just the compilation of Nim and tools?

heinthanth:
Sure, I'll

heinthanth:
the problem here is embedded `proxyexe` is x86 binary. That's why binaries in `~/.nimble/bin` are x86 although actually compiled toolchain in `~/.choosenim/toolchains/version/bin/` are arm64. So, we need to fetch `cliparams.nim`, `common.nim`, `proxyexe.nim` and re-compile `proxyexe` with recently built arm nim binary ( if first installed ).

heinthanth:
Analytics not working on M1. Seems like segment fault from `objc_msgSend`

Screen Shot 2022-01-27 at 11 05 16 AM

heinthanth:
Now, until puppy's objc binding is fixed for M1, I passed `-d:puppyLibcurl` for Apple Silicon to use libcurl instead of AppKit. Now, we have universal macos binary for both intel and M1. I add cross-compilation and universal binary stuffs on scripts/build.sh. Please kindly check it out @dom96

heinthanth:
Travis is failing .. don't know why ... sorry to throw **comment bomb** into this PR.

heinthanth:
Force pushed ... since I think I messed up. Everything is working file on both intel and m1 mac. But travis is Failing ... May be I can figure out travis tonight.

heinthanth:
For now, I will close this PR. Since I'm not OK with travis ( ran out of credit - accidentally enable Consume paid credits for OSS ), I'll continue my fork with Github Action. And reopen soon! Thanks @dom96 for your encouragement.

> This is a cool approach, thank you!

dom96:
> the problem here is embedded proxyexe is x86 binary. That's why binaries in ~/.nimble/bin are x86 although actually compiled toolchain in ~/.choosenim/toolchains/version/bin/ are arm64.

Why is this a problem btw? all that matters is that the Nim binaries are arm64, no?

heinthanth:
No ... proxyexe is x86 and if we run nim c something.nim, something  also become x86.

dom96:
proxyexe just executes the arm64 Nim though, or am I missing something? why would proxyexe being arm64 matter in what `nim c ...` produces?

heinthanth:
I've demo here:

I create a simple proxy which invoke my arm64 nim binary.

Screen Shot 2022-01-28 at 1 54 31 PM

Here, I have two nim binaries. I compiled proxy.nim with x86 nim.

Screen Shot 2022-01-28 at 1 54 46 PM

So, proxy binary become x86.

Screen Shot 2022-01-28 at 2 33 15 PM

Then, I tried to compile `test.nim` with proxy binary ( which, in turn, invoke arm64 nim binary ). But test binary become x86.

Screen Shot 2022-01-28 at 1 56 28 PM

I think, test.nim also become x86 since the parent process which invoke C compile is x86. Don't know much about nim internal stuffs ( how nim invoke c compiler ).

But when I compile test.nim with my nim arm64 binary, it become ARM.

Screen Shot 2022-01-28 at 1 57 09 PM

I think choosenim has two options:

* use another symlink method instead of proxy binary.
* cross-compile choosenim to ARM64 on x86_64 macOS.

If we want, we can discord too. We're friends on discord.

PR #294, opened 2022-01-29
build fat binaries on macosx

elcritch:
It looks like choosenim is still using rosetta x86 builds for M1 based Macs. This change _should_ get the macosx build to generate "fat binaries" that support both x86_64 and arm64 variants. 

I'm testing it on my M1, but don't have access to an x86 Mac to verify the binaries work there as well.

### DIFF
--- a/src/choosenim.nims
+++ b/src/choosenim.nims
@@ -1,5 +1,7 @@
 when defined(macosx):
   switch("define", "curl")
+  switch("passC","-arch arm64 -arch x86_64")
+  switch("passL","-arch arm64 -arch x86_64")
 elif not defined(windows):
   switch("define", "ssl")
 


heinthanth:
Yes ... it's good!

heinthanth (approved):

dom96:
Epic. I've got a x86 mac so I should be able to test this :)

elcritch:
Great! I tried getting Nim itself to compile a fat binary too. But, there's too many levels of bootstrap builds. It did produce arm64 native builds. Any thoughts on which stage I'd need to modify to get the Nim compiler to build a fat binary?

heinthanth:
Hello, will u merge this? We have to manual compile and install nim from source till now.

dom96:
I suppose unsurprisingly I cannot build with these flags on a x86 Mac. So I cannot test it. But I will merge and we can see if @marcomq's build script succeeds in building binaries for this.

dom96:
Sadly not:

image

heinthanth:
I think it's due to Zippy dependency which use `asm` on `amd64`. I'm figuring out.

elcritch:
Thanks! Though making choosenim a fat binary doesn't fix the compiler itself. That might require setting up choosenim to download a different Nim binary when it's run on MacOS/ARM.

heinthanth:
Yeah ... 
1. when cross-compiling arm64 from amd64 with `nim c --passC:'-arch arm64' --passL:'-arch arm64'`,
2. `amd64` is defined and this lines are activated: https://github.com/guzba/zippy/blob/8e472c468c1c67b3221a71ad7224cd058492b752/src/zippy/internal.nim#L323-L336
3. To fix that we can set `--cpu` flag.

To fix that, we can separate the build and then, can create fat bin with `lipo`. What do u think?
Choosenim also use proxy binaries. So, we can't build universal bin. Instead, we have to use separate choosenim arm64 and amd64.

dom96:
Personally I think that implementing arm support for choosenim is far more important than making choosenim itself a fat binary. As long as macOS will support Rosetta it should be fine. 

What this ideally means is also arm Linux support. But this is trickier. If we can get choosenim to build an arm macOS Nim that would be a great first step.

heinthanth:
Hello ... I've created an initial patch that works: https://github.com/dom96/choosenim/pull/301. I think I need to M1 user to test this 🤔 .

elcritch:
@heinthanth awesome! I'll test it out here and see if it works.

PR #296, opened 2022-03-21
Adding workflow to automatically build and upload binaries

marcomq:
This workflow would automatically build and upload binaries for windows, linux and macos when tags are created. It may be triggered manually and would overwrite existing binaries.
On linux, it would use manylinux2014, so the binary would probably be compatible for RHEL 7 and other older and newer distributions.

Note: For windows, this would currently just upload the .exe and not the .bat /.zip file. Couldn't find the sources for those.

I only tested the manual trigger - not sure if upload on create tag would do everything correct.
You may delete this if you currently have some other solution :)

### DIFF
--- a/.github/workflows/release_build.yml
+++ b/.github/workflows/release_build.yml
@@ -26,9 +26,6 @@ jobs:
       - uses: actions/checkout@v2
       - name: Build binary
         run: |
-          echo `pwd`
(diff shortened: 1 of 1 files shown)


dom96 (approved):
Nice, let's give it a try

PR #304, opened 2022-08-04
Add documentation to help mac users

lantos1618:
Mac users need to install rosetta and xcode
both of the following need to be installed
```
xcode-select --install
softwareupdate --install-rosetta
```
otherwise new users will be faced with
```
choosenim-init: Downloading choosenim-0.8.4_macosx_amd64
sh: line 77: /var/folders/md/61b29s79495glw9mcnyhmbjr0000gn/T//choosenim-0.8.4_macosx_amd64: Bad CPU type in executable
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -49,6 +49,18 @@ curl https://nim-lang.org/choosenim/init.sh -sSf | sh
 wget -qO - https://nim-lang.org/choosenim/init.sh | sh
 ```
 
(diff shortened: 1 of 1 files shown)


dom96 (commented):
Thank you for this! Ideally we'd have the init.sh script check for this and give a nice error, any chance you'd be willing to implement that? :)

PR #341, opened 2024-04-03
Updated readme to address potential issue with MacOs

Niminem:
I banged my head against the wall for a couple of hours trying to get nim/choosenim/etc to work on my mac- I clearly saw `~/.nimble/bin` in my path via `echo $path` but it wouldn't work. Upon further investigation, apparently I didn't have a .zshrc file in my home directory and needed to add the path there.

This was on a clean installation of mac.

It may or may not be common so I figured it's best to add notes to the README.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -78,6 +78,14 @@ wget -qO - https://nim-lang.org/choosenim/init.sh | sh
               script to install by specifying the ``CHOOSENIM_CHOOSE_VERSION``
               environment variable.
 
+**MacOs Note** If `choosenim` has been installed successfully, and you're still not able to use Nim or any of the tooling (ex: `nim --v`) follow these steps in your terminal:
+- modify or create the `~/.zshrc` file with this line: `nano ~/.zshrc`
+- add the following line to setup the PATH correctly: `export PATH="$HOME/.nimble/bin:$PATH"`
+- make sure the current terminal session picks up the changes with this line: `source ~/.zshrc`
+- now validate everything works properly with: `nim --v`
+
+By creating a `.zshrc` file and adding the path to `~/.nimble/bin`, you're ensuring that zsh (terminal) knows where to look for your Nim executables and other tooling every time it starts up.
+
 ## How choosenim works
 
 Similar to the likes of ``rustup`` and ``pyenv``, ``choosenim`` is a
--- a/readme.md
+++ b/readme.md
@@ -78,14 +78,14 @@ wget -qO - https://nim-lang.org/choosenim/init.sh | sh
               script to install by specifying the ``CHOOSENIM_CHOOSE_VERSION``
               environment variable.
 
-**MacOs Note** If `choosenim` has been installed successfully, and you're still not able to use Nim or any of the tooling (ex: `nim --v`) follow these steps in your terminal:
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -78,6 +78,14 @@ wget -qO - https://nim-lang.org/choosenim/init.sh | sh
...
               environment variable.
 
+**MacOs Note** If `choosenim` has been installed successfully, and you're still not able to use Nim or any of the tooling (ex: `nim --v`) follow these steps in your terminal:
+- modify or create the `~/.zshrc` file with this line: `nano ~/.zshrc`
+- add the following line to setup the PATH correctly: `export PATH="$HOME/.nimble/bin:$PATH"`
+- make sure the current terminal session picks up the changes with this line: `source ~/.zshrc`
+- now validate everything works properly with: `nim --v`
+
+By creating a `.zshrc` file and adding the path to `~/.nimble/bin`, you're ensuring that zsh (terminal) knows where to look for your Nim executables and other tooling every time it starts up.
+
```
janAkali: Good idea, but maybe it should be a note for all unix-like systems. It's already 80% there - zsh is default shell on mac, but linux distros most commonly use bash. Everything else in your note applies to both systems.

```suggestion
**PATH setup** If `choosenim` has been installed successfully, and you're still not able to use Nim or any of the tooling (ex: `nim --v`) follow these steps in your terminal:
- MacOs: modify or create the `~/.zshenv` file with your preferred editor, e.g. `nano ~/.zshenv`
- Linux: modify or create file `~/.bash_profile`, e.g. `nano ~/.bash_profile`
- add the following line to setup the PATH correctly: `export PATH="$HOME/.nimble/bin:$PATH"`
- make sure the current terminal session picks up the changes with this line: `source ~/.zshenv` or `source ~/.bash_profile`
- now validate everything works properly with: `nim --v`

By creating a `.zshenv` or `.bash_profile` file and adding the path to `~/.nimble/bin`, you're ensuring that your shell (terminal) knows where to look for your Nim executables and other tooling every time it starts up.
```
The proper place for environment variables is `~/.zshenv` for zsh and `~/.bash_profile` for bash.
These configs are executed on a system startup instead of `.zshrc` which is run when you open a terminal - so GUI apps like VScode could find nim executable.

Here's a good reference on a PATH env variable for MacOS: https://gist.github.com/Linerre/f11ad4a6a934dcf01ee8415c9457e7b2

janAkali:
> I clearly saw ~/.nimble/bin in my path via echo $path but it wouldn't work.

$path and $PATH are two different things. $PATH is widely used by everything, but I've never seen $path until today.

Niminem:
@Archargelod Interesting. Oddly enough, both `$PATH` and `$path` shown the same for me and figured it was just case-insensitive and the same things. I've accepted those changes from your end. I think we may be good here to finish this one out :)

Niminem:
@Archargelod @dom96 Gentlemen, I would love to make some Youtube videos about Nim and simply point them to Choosenim for installation. Because many of the viewers will not be from a technical background (I do SEO professionally), many will have a Mac and may encounter the issue described above.

Can we merge this PR asap?

janAkali (approved):

Niminem:
@dom96 ?

